PR #41, opened 2026-04-09
ci: add IDL freshness check and consolidate artifacts

0x-r4bbit:
Move IDL files to artifacts/ and add a convention-based CI check that discovers all programs via */methods/guest/src/bin/*.rs and fails if any program is missing its IDL or has one that is out of date.


.github/workflows/ci.yml
```diff
@@ -116,3 +116,48 @@ jobs:
...
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
+
+      - name: Build idl-gen
+        run: cargo build -p idl-gen --release
```
0x-r4bbit, line 143: Notice that this builds a custom idl-gen instead of relying on SPEL CLI.

The reason is that installing SPEL is tricky as it ultimately depends on logos-circuits, which are private.
But we don't need those for IDL generation, so `idl-gen` uses the SPEL library create to do the job (instead of cargo install'ing spel cli).

3esmit (changes requested):
Looks good to me. Only missing optional update on documentation.

Now that CI is checking IDL freshness through the repo-local `idl-gen` binary, I think the docs should point contributors at that same command as the canonical regeneration path instead of `spel generate-idl`.

`spel` may be equivalent today, but it is still an external tool/version boundary, while CI is pinning behavior through `cargo run -p idl-gen -- <path>`. Documenting the repo-local command here would make it clearer which workflow is authoritative, and `spel generate-idl` can still be mentioned as an equivalent convenience if you want to keep that note.

The docs to update are README.md (line 75) and CLAUDE.md (line 36).

0x-r4bbit:
@3esmit addressed your comment.

PR #45, opened 2026-04-11
fix(amm): implement k-invariant check in swap post-state creation and…

### LINKED ISSUE #5: Enforce post-swap k invariant in AMM program
The work was already done here https://github.com/logos-blockchain/logos-execution-zone/pull/373

We need to move this PR over.

3esmit:
## 🎯 Purpose
Fixes #5, replaces #14.
This PR rebases the branch onto current `main` and adds a post-swap `k` invariant check to the AMM swap flow.

## ⚙️ What Changed
- [x] Rebase the branch onto current `main` and resolve the swap changes against the current `swap_exact_input` / `swap_exact_output` layout.
- [x] Add a widened `k` comparison inside `amm/src/swap.rs`'s shared `create_swap_post_states` helper, so the invariant is checked at the same point reserves are written back to the pool definition.
- [x] Because both `swap_exact_input` and `swap_exact_output` route through that helper, the post-swap invariant is now checked for both swap modes.
- [x] Use `u128::carrying_mul` for the widened comparison so the invariant check remains correct when `reserve_a * reserve_b` exceeds `u128`.
- [x] Add focused AMM test helpers and a widened-`k` regression test that exercises the shared post-state helper.
- [x] Leave public AMM interfaces unchanged; no guest, core, or IDL updates are required.

## 🧪 How to Test
1. `RISC0_DEV_MODE=1 cargo +1.94.0 test -p amm_program --lib swap`
2. `RISC0_DEV_MODE=1 cargo +1.94.0 test --workspace --exclude integration_tests`
3. `RISC0_DEV_MODE=1 cargo +1.94.0 test -p integration_tests`

### DIFF
--- a/amm/src/swap.rs
+++ b/amm/src/swap.rs
@@ -76,20 +76,34 @@ fn create_swap_post_states(
     deposit_b: u128,
     withdraw_b: u128,
 ) -> Vec<AccountPostState> {
+    let new_reserve_a = pool_def_data
+        .reserve_a
+        .checked_add(deposit_a)
+        .expect("reserve_a + deposit_a overflows u128")
+        .checked_sub(withdraw_a)
+        .expect("reserve_a + deposit_a - withdraw_a underflows");
+    let new_reserve_b = pool_def_data
+        .reserve_b
+        .checked_add(deposit_b)
+        .expect("reserve_b + deposit_b overflows u128")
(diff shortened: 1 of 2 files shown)


0x-r4bbit (commented):
As discussed previously offline, this turns out not to be very useful here.
There's no case in which `new_k < old_k`.

PR #21, opened 2026-04-01
feat(amm): apply trading fees to LP accounting

3esmit:
## Purpose
Build on the #20 stage-1 fee-tier work by applying trading fees to swap and liquidity accounting so LPs earn and realize fees through normal AMM usage. 
Closes #19 

## What Changed
- Cherry-picked the surplus-removal fix from #16 so this branch includes the live-vault withdrawal behavior required for LP fee realization.
- Applied fee-aware swap pricing using the pool's configured fee tier while continuing to transfer the full input amount into the vault.
- Updated tracked reserves with the fee-adjusted input amount so the charged fee remains in the vault as LP-owned surplus.
- Updated `add_liquidity` to mint LP against live vault balances when surplus exists, preventing new LPs from receiving fees accrued before they joined.
- Kept `remove_liquidity` reserve accounting intact while distributing proportional live vault balances, allowing withdrawing LPs to realize accrued trading fees.
- Added AMM unit and integration coverage for fee rounding edge cases, repeated swaps, fee payout on liquidity removal, and add-liquidity behavior after fees accrue.

## Review Focus
- Confirm the cherry-picked surplus-removal change from `fix/remove-lp-surplus-missing` is the intended base for this PR.
- Swap reserve updates now intentionally differ from vault balance changes by the configured fee amount.
- Live vault balances now affect LP minting when surplus exists.
- Liquidity removal should pay out accumulated surplus proportionally without mutating tracked reserves incorrectly.
- Multi-transaction integration scenarios now rely on current account nonces as fees accumulate across sequential operations.

## How To Test
1. `cargo +nightly fmt --all -- --check`
2. `taplo fmt --check .`
3. `RISC0_SKIP_BUILD=1 cargo +1.94.0 clippy --workspace --all-targets -- -D warnings`
4. `RISC0_DEV_MODE=1 cargo +1.94.0 test --workspace --exclude integration_tests`
5. `RISC0_DEV_MODE=1 cargo +1.94.0 test -p integration_tests`

Issue #19, opened 2026-03-31
AMM: Implement fee system - Trading Fees For LPs


3esmit:
### Scope
- Apply the fee only in swap pricing, not by changing the transferred input amount.
- Leave the fee amount in the vault as surplus above tracked reserves.
- Continue distributing vault balances proportionally on liquidity removal so accrued fees are claimable by LPs.
- Update liquidity-add math to use live vault balances when surplus exists so newly added LPs do not receive a share of fees earned before they joined.

### Acceptance Criteria
- Repeated swaps accumulate measurable fee surplus in the vaults.
- Removing liquidity pays out a proportional share of the accrued surplus.
- Adding liquidity after fee accrual does not dilute previously earned LP fees.
- Unit and integration coverage proves multi-swap fee accumulation and post-fee

Implement trading fee accumulation so that liquidity providers earn a portion of each swap based on the pool's configured fee tier.
Fees accumulate as surplus in the pool vaults above the tracked reserves and are distributed to LPs proportionally when they remove liquidity.

### Dependency

Depends on #15

0x-r4bbit:
This has landed via #46

Issue #7, opened 2026-03-30
AMM: Implement fee system


0x-r4bbit:
## Summary
Implement the AMM fee system in two connected stages. Stage 1 introduces configurable fee tiers at pool creation and locks the fee selection into pool state. Stage 2 makes swap fees accrue to liquidity providers and preserves that accrued value when new liquidity is added later.
According to what's described here: https://github.com/logos-co/rfp/blob/master/RFPs/RFP-004-privacy-preserving-dex.md

## Background
The AMM already stores a `fees` field in `PoolDefinition`, but the field is unused today. This work wires that field into the public instruction surface and then applies it to swap accounting so LPs earn fees through normal pool usage.

## Dependency (Stage 0): #15 
`remove_liquidity` currently pays out against tracked reserves (`reserve_a` and `reserve_b`) instead of the live token balances held in the vault accounts. Its `delta_lp` expression is also an identity that always collapses to the raw `remove_liquidity_amount`, so LP burn semantics already behave as "burn exactly what the caller requested." Once trading fees accumulate as surplus in the vaults, LP exits should receive their proportional share of the actual vault balances, not just the tracked reserves.

## Stage 1: #18 
Implement configurable fee tiers for the AMM supporting 0.01%, 0.05%, 0.3%, and 1% fee rates as required by RFP-004.
The `fees: u128` field already exists in `PoolDefinition` but is unused. Wire it up so pool creators can select a fee tier at pool creation time.

## Stage 2: #19
Implement trading fee accumulation so that liquidity providers earn a portion of each swap based on the pool's configured fee tier.
Fees accumulate as surplus in the pool vaults above the tracked reserves and are distributed to LPs proportionally when they remove liquidity.

## Why The Stages Are Split
Stage 1 changes the public interface and persisted pool configuration. Stage 2 depends on that stored configuration to apply fee-aware swap math and LP accounting. Splitting the work keeps the API/state change reviewable on its own while preserving a direct path to the full fee system.

0x-r4bbit:
This has landed via #46

PR #46, opened 2026-04-13
feat(amm): apply trading fees to LP accounting

0x-r4bbit:
Implement Uniswap V2-style fees-in-reserves: the full swap_amount_in is deposited into the reserve (growing k = reserve_a * reserve_b), while only the fee-adjusted effective_amount_in is used to compute the output amount. This means LPs earn fees proportionally on every removal via k-growth rather than through a separate vault surplus.

- swap_logic: add fee_bps parameter; compute effective_amount_in for output formula only; return full swap_amount_in as the reserve deposit
- add_liquidity: remove vault balance checks; use pool reserves directly for ratio and LP calculations (vault == reserve invariant holds)
- Fix all integration test fixture values to match fees-in-reserves math
- Remove dead-code vault_a/b_init_zero helpers from unit tests

### DIFF
--- a/amm/src/add.rs
+++ b/amm/src/add.rs
@@ -1,6 +1,9 @@
 use std::num::NonZeroU128;
 
-use amm_core::{assert_supported_fee_tier, compute_liquidity_token_pda_seed, PoolDefinition};
+use amm_core::{
+    assert_supported_fee_tier, compute_liquidity_token_pda_seed, read_vault_fungible_balances,
+    PoolDefinition,
+};
 use nssa_core::{
     account::{AccountWithMetadata, Data},
     program::{AccountPostState, ChainedCall},
@@ -44,33 +47,9 @@ pub fn add_liquidity(
         "Both max-balances must be nonzero"
     );
 
-    // 2. Determine deposit amount
-    let vault_b_token_holding = token_core::TokenHolding::try_from(&vault_b.account.data)
-        .expect("Add liquidity: AMM Program expects valid Token Holding Account for Vault B");
-    let token_core::TokenHolding::Fungible {
-        definition_id: _,
-        balance: vault_b_balance,
-    } = vault_b_token_holding
-    else {
-        panic!(
-            "Add liquidity: AMM Program expects valid Fungible Token Holding Account for Vault B"
-        );
-    };
+    let (vault_a_balance, vault_b_balance) =
+        read_vault_fungible_balances("Add liquidity", &vault_a, &vault_b);
 
-    let vault_a_token_holding = token_core::TokenHolding::try_from(&vault_a.account.data)
-        .expect("Add liquidity: AMM Program expects valid Token Holding Account for Vault A");
-    let token_core::TokenHolding::Fungible {
-        definition_id: _,
-        balance: vault_a_balance,
-    } = vault_a_token_holding
-    else {
-        panic!(
-            "Add liquidity: AMM Program expects valid Fungible Token Holding Account for Vault A"
-        );
-    };
-
-    assert!(pool_def_data.reserve_a != 0, "Reserves must be nonzero");
-    assert!(pool_def_data.reserve_b != 0, "Reserves must be nonzero");
     assert!(
         vault_a_balance >= pool_def_data.reserve_a,
         "Vaults' balances must be at least the reserve amounts"
@@ -80,7 +59,10 @@ pub fn add_liquidity(
         "Vaults' balances must be at least the reserve amounts"
     );
 
-    // Calculate actual_amounts
+    // 2. Determine deposit amount
+    assert!(pool_def_data.reserve_a != 0, "Reserves must be nonzero");
+    assert!(pool_def_data.reserve_b != 0, "Reserves must be nonzero");
+
     let ideal_a: u128 = pool_def_data
         .reserve_a
         .checked_mul(max_amount_to_add_token_b)
--- a/amm/src/swap.rs
+++ b/amm/src/swap.rs
@@ -1,4 +1,6 @@
-use amm_core::{assert_supported_fee_tier, MINIMUM_LIQUIDITY};
+use amm_core::{
+    assert_supported_fee_tier, read_vault_fungible_balances, FEE_BPS_DENOMINATOR, MINIMUM_LIQUIDITY,
+};
 pub use amm_core::{compute_liquidity_token_pda_seed, compute_vault_pda_seed, PoolDefinition};
 use nssa_core::{
     account::{AccountId, AccountWithMetadata, Data},
@@ -28,31 +30,13 @@ fn validate_swap_setup(
         "Vault B was not provided"
     );
 
-    let vault_a_token_holding = token_core::TokenHolding::try_from(&vault_a.account.data)
-        .expect("AMM Program expects a valid Token Holding Account for Vault A");
-    let token_core::TokenHolding::Fungible {
-        definition_id: _,
-        balance: vault_a_balance,
-    } = vault_a_token_holding
-    else {
-        panic!("AMM Program expects a valid Fungible Token Holding Account for Vault A");
-    };
+    let (vault_a_balance, vault_b_balance) =
+        read_vault_fungible_balances("Validate swap setup", vault_a, vault_b);
 
     assert!(
         vault_a_balance >= pool_def_data.reserve_a,
         "Reserve for Token A exceeds vault balance"
     );
-
-    let vault_b_token_holding = token_core::TokenHolding::try_from(&vault_b.account.data)
-        .expect("AMM Program expects a valid Token Holding Account for Vault B");
-    let token_core::TokenHolding::Fungible {
-        definition_id: _,
-        balance: vault_b_balance,
-    } = vault_b_token_holding
-    else {
-        panic!("AMM Program expects a valid Fungible Token Holding Account for Vault B");
-    };
-
     assert!(
         vault_b_balance >= pool_def_data.reserve_b,
         "Reserve for Token B exceeds vault balance"
@@ -130,6 +114,7 @@ pub fn swap_exact_input(
                 user_holding_b.clone(),
                 swap_amount_in,
                 min_amount_out,
+                pool_def_data.fees,
                 pool_def_data.reserve_a,
                 pool_def_data.reserve_b,
                 pool.account_id,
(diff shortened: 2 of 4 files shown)


amm/src/add.rs
```diff
@@ -45,42 +47,9 @@ pub fn add_liquidity(
...
     assert!(pool_def_data.reserve_a != 0, "Reserves must be nonzero");
     assert!(pool_def_data.reserve_b != 0, "Reserves must be nonzero");
-    assert!(
-        vault_a_balance >= pool_def_data.reserve_a,
-        "Vaults' balances must be at least the reserve amounts"
-    );
-    assert!(
-        vault_b_balance >= pool_def_data.reserve_b,
-        "Vaults' balances must be at least the reserve amounts"
-    );
```
0x-r4bbit, line 81: While this is an invariant that must be true at all times, I don't think there's a point in having this checked every time someone wants to add liquidity.

amm/src/add.rs
```diff
@@ -45,42 +47,9 @@ pub fn add_liquidity(
...
-        .expect("Add liquidity: AMM Program expects valid Token Holding Account for Vault A");
-    let token_core::TokenHolding::Fungible {
-        definition_id: _,
-        balance: vault_a_balance,
-    } = vault_a_token_holding
-    else {
-        panic!(
-            "Add liquidity: AMM Program expects valid Fungible Token Holding Account for Vault A"
-        );
-    };
```
0x-r4bbit, line 70: LP calculation should be done on the reserves only (assuming they are up-to-date)
"Not up-to-date" means there's been a donation in the meantime. In which case, one can call `syncReserves` first.

We might also want to look into syncing reserves on any state changing operation to reduce likelihood of surplus.

amm/src/add.rs
0x-r4bbit: On a second thought, if **anything goes wrong** and for some reason reserves are indeed greater than the token balance, it might actually be good to prevent users from adding liquidity.

Might revert this change. Let me know what you guys think.

amm/src/tests.rs
```diff
@@ -683,37 +694,6 @@ impl AccountWithMetadataForTests {
...
-                data: Data::from(&TokenHolding::Fungible {
-                    definition_id: IdForTests::token_b_definition_id(),
-                    balance: 0,
-                }),
-                nonce: Nonce(0),
-            },
-            is_authorized: true,
-            account_id: IdForTests::vault_b_id(),
-        }
-    }
```
0x-r4bbit, line 717: Removing cause they aren't used anywhere

amm/src/tests.rs
```diff
@@ -1431,40 +1435,6 @@ fn test_call_add_liquidity_zero_balance_2() {
...
-        AccountWithMetadataForTests::vault_b_init_zero(),
-        AccountWithMetadataForTests::pool_lp_init(),
-        AccountWithMetadataForTests::user_holding_a(),
-        AccountWithMetadataForTests::user_holding_b(),
-        AccountWithMetadataForTests::user_holding_lp_init(),
-        NonZero::new(BalanceForTests::add_max_amount_a()).unwrap(),
-        BalanceForTests::add_max_amount_b(),
-        BalanceForTests::add_min_amount_lp(),
-    );
-}
```
0x-r4bbit, line 1377: Removed these because I've removed the assertions mentioned above.
Depending on decision, I can reintroduce them.

amm/src/add.rs
3esmit: Even if `vault_balance >= reserve` is supposed to be an invariant, this is still a useful fail-closed check when the pool is already undercollateralized or otherwise corrupted. Without it, `add_liquidity` can keep minting LP and advance reserve accounting against assets that are not actually present in the vaults.

And I agree with the second take here: If reserves ever exceed the live vault balance, letting `add_liquidity` continue would hide a bad state and mint LP against nonexistent assets. I would revert this part of the change and keep the explicit guard in place.

amm/src/add.rs
3esmit: I think this is the main spec mismatch in the PR.

Issue #19 and the parent issue #7 explicitly describe a surplus-based model: the fee stays in the vault above tracked reserves, remove-liquidity pays from actual vault balances, and add-liquidity should use live vault balances after fee accrual so new LPs do not inherit previously earned fees. If we want reserve-only math instead, I think we should rewrite the issue/acceptance criteria first and review that design change explicitly. 

Do you want me to do that?

amm/src/tests.rs
3esmit: This makes sense only if we are sure the add-liquidity undercollateralization check is gone for good.

If we keep or restore that guard, I would also restore these helpers or replace them with equivalent fixtures, because we still want focused coverage for the `vault < reserve` failure path.

amm/src/tests.rs
3esmit: Same here.

If the undercollateralization guard in `add_liquidity` stays, I would bring these tests back or replace them with equivalent coverage. Right now the PR removes the guard and the dedicated regression tests together, which makes it harder to spot this fail-closed behavior changing.

3esmit (changes requested):
Revert removal of add_liquidity guard

amm/src/add.rs
0x-r4bbit: You can write the issue if you want. @fryorcraken has there been a verdict on whether or not we allow for recover surplus? My understanding was that we don't want/need it, so any surplus can be synced to the pool reserves (which benefits liquidity providers).

LPs are then always calculated based on reserves.
If a surplus has been created before and the LP doesn't call sync before removing, then it would be on them.

amm/src/add.rs
0x-r4bbit: Reintroduced these assertions

amm/src/tests.rs
0x-r4bbit: Reintroduced

gravityblast (approved):

amm/src/swap.rs
```diff
@@ -318,12 +316,24 @@ fn exact_output_swap_logic(
         "Exact amount out exceeds reserve"
     );
 
-    // Compute deposit amount using ceiling division
-    // Formula: amount_in = ceil(reserve_in * exact_amount_out / (reserve_out - exact_amount_out))
-    let deposit_amount = reserve_deposit_vault_amount
+    // Compute the gross deposit amount the user must pay (fee inclusive).
```
3esmit: Change the exact-output input formula. 

```rust
effective_in_min = ceil(reserve_in * amount_out / (reserve_out - amount_out))
deposit_amount = ceil(effective_in_min * FEE_DENOM / (FEE_DENOM - fee))
```

This matches the exact-input rounding and stops the 1-token undercharge case.

3esmit (dismissed):

3esmit:
The issue is that `exact_output_swap_logic` is supposed to invert the exact-input pricing rule, but it inverts the continuous formula instead of the discrete integer-rounded one.

In exact input  
https://github.com/logos-blockchain/lez-programs/blob/6d057aeda585aa52d76aa0f60f216ca9e04d95bb/amm/src/swap.rs#L172-L188

the executed pricing function is:

```rust
effective_amount_in = floor(input * (10000 - fee) / 10000)
withdraw_amount = floor(reserve_out * effective_amount_in / (reserve_in + effective_amount_in))
```

In exact output  
https://github.com/logos-blockchain/lez-programs/blob/6d057aeda585aa52d76aa0f60f216ca9e04d95bb/amm/src/swap.rs#L328-L336

the code computes:

```rust
deposit_amount = ceil(reserve_in * amount_out * 10000 / ((reserve_out - amount_out) * (10000 - fee)))
```

That `ceil` is the inverse of the ideal rational formula, not the inverse of the actual integer-rounded function above. Because the exact-input path floors twice, `deposit_amount` can be 1 too small.

The proof test  
https://github.com/logos-blockchain/lez-programs/blob/6d057aeda585aa52d76aa0f60f216ca9e04d95bb/amm/src/tests.rs#L2442-L2483

shows the concrete case:
- pool `1000 / 1000`, fee `30 bps`, target output `1`
- exact-output accepts gross input `2`
- but exact-input math with gross input `2` gives `effective_in = 1`, then `amount_out = 0`
- the true minimum gross input is `3`

So the bug is: **exact-output computes `ceil(ideal inverse)`, but exact-input executes a floored integer function, and those are not equivalent**.

3esmit:
I am bringing back part of the changes from #21, specifically
https://github.com/3esmit/lez-programs/blob/d634f05ede972c0de4fb9aca49fd37531c8d9857/amm/src/swap.rs#L335-L351

3esmit:
`fix(swap): revert changes in exact_output_swap_logic from original PR` actually fixes the rounding error, the reason the tests failed are because of `call_swap_exact_output_chained_call_successful` which was expecting different format. 
https://github.com/logos-blockchain/lez-programs/actions/runs/24416362575/job/71326484229#step:6:313
`test tests::call_swap_exact_output_rejects_max_in_that_rounds_down_below_target_output - should panic ... ok`

While fixing the test I realized that the format introduced by @0x-r4bbit, where the rounding happened was more clear and maintainable, so instead of changing the test, I adapted the division to apply that same format where numerator and denominator are separated. I will now request a review from AI and fix any findings they encounter.

amm/src/add.rs
```diff
@@ -80,7 +59,10 @@ pub fn add_liquidity(
         "Vaults' balances must be at least the reserve amounts"
     );
```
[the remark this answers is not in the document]
3esmit: @0x-r4bbit Can you answer this one?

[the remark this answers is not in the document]
0x-r4bbit: Ah, I need to update the commit message. This is about the invariant checks re: token.balance >= vault.reserves that I've originally removed from add_liquidity (which we then decided to put back in)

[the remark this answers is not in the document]
0x-r4bbit: Updated the commit message

0x-r4bbit:
@3esmit Rebased and squashed your changes in.

Thank you for the fix! Merging this now

amm/src/add.rs
fryorcraken: I suggest no recover surplus, in line with your expectatiojns: https://github.com/logos-co/rfp/pull/41

PR #47, opened 2026-04-13
chore(amm): new_definition allows only uninitialized pools

gravityblast:
This PR updates `amm/new_definition` to accept only uninitialized pools.

closes #26


0x-r4bbit:
@gravityblast can you make sure this passes on CI?

0x-r4bbit (approved):
Changes LGTM but need to pass on CI.

Also there's #44 which likely has to land first.

PR #52, opened 2026-04-15
Chore/update lez

### LINKED ISSUE #43: Update to LEZ dependency to latest version
The latest version of LEZ introduces a bunch of very useful changes:

- Validity windows (PRs [400](https://github.com/logos-blockchain/logos-execution-zone/pull/400), [404](https://github.com/logos-blockchain/logos-execution-zone/pull/404)). Now program outputs have extra fields to set time and block based windows where executions are valid.
- PDA griefing attack protection (PR [414](https://github.com/logos-blockchain/logos-execution-zone/pull/414)). Public account initialization requires is_authorized=true now. This comes with the drawback that public recipients in token transfers need to be previously initialized (no initialization on the fly anymore)
- Self program id input (PR [426](https://github.com/logos-blockchain/logos-execution-zone/pull/426)). Programs input now include a new self_program_idargument that's always populated with the exeucting program id. So now programs can check if an account is owned by them by comparing the program_owner to the self_program_id.
- Caller program id (PR [428](https://github.com/logos-blockchain/logos-execution-zone/pull/428)). Program input now includes a new caller_program_id argument that's populated with the caller in a chain call. This can be used for example by program to define functions that can only be executed if called from a specific program. Check out the flash loan example where this is used to define an "internal" function that can only be called as a chain call from the same program.
- Clock accounts (PR [403](https://github.com/logos-blockchain/logos-execution-zone/pull/403)). The state now is initialized with a set of accounts that carry block and timestamp information. Programs can use these as input to get block context information if they need it. Check out the timestamp-based piñata example, where these are used to prevent abuse by setting a cooldown period

Specifically #8 depends on the block context. 
So to unblock #8 we need to update our LEZ dependency first.

### LINKED ISSUE #48: Update SPEL to the latest version
This likely needs to be done together with #43

3esmit:
Closes #43
Closes #48

## Summary

This PR updates LEZ to `v0.2.0-rc1` and SPEL to `v0.2.0-rc2`.  That also required updating the direct SPEL dependencies to the current RC line, because the guest entrypoints and IDL tooling in this repo consume the LEZ runtime through `spel-framework`.

It also started to use `tag` instead of `rev` because SPEL is using `tag`, but the Cargo.lock is a tracked file, and as it carries the exact commit hash, the safety result ends up being identical. 

The main LEZ changes that matter here are the new public-claim rules, plus the new `self_program_id` and `caller_program_id` plumbing. The Token, AMM, and ATA changes in this PR were chosen to keep the existing programs stable under those upstream rules, and the implementation follows the decisions discussed in LEZ PRs.

Upstream context the changes are within the main LEZ discussion in:

https://github.com/logos-blockchain/logos-execution-zone/pull/414

https://github.com/logos-blockchain/logos-execution-zone/pull/426

https://github.com/logos-blockchain/logos-execution-zone/pull/428

## Justification

LEZ changed how public accounts can be claimed. A public account can now only be claimed if the call proves authorization for that account or proves a PDA seed for it. That affects how Token creates new public holdings, how AMM creates the first LP holding for a user, and how ATA transfers into a fresh recipient ATA.

The LEZ update also changed the runtime input shape through `self_program_id` and `caller_program_id`. Those changes are mostly absorbed through the dependency update, but they are part of why LEZ and SPEL had to move together in this PR.

## Important contract changes

- Token public mint and transfer no longer assume that a raw public recipient can always be created on the fly. A fresh public recipient now needs explicit authorization in the same transaction.
- `AMM::NewDefinition` now requires `user_lp` authorization when the LP holding is being created for the first time.
- `ATA::Transfer` is now explicitly ATA-to-ATA and takes `(owner, sender_ata, recipient_owner, recipient_ata)`.

### `b1a5afb` refactor!: Update dependencies and implement new required features across multiple modules

This is the base compatibility commit. It aligns direct LEZ pins to `v0.2.0-rc1`, aligns direct SPEL pins to `v0.2.0-rc.2`, refreshes the lockfiles, and updates the claim and post-state handling needed by the LEZ bump across Token, AMM, and ATA. This is a breaking compatibility change because it updates the runtime contract used by all three programs.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/Cargo.toml#L22-L24

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/token/methods/guest/Cargo.toml#L12-L19

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/ata/src/create.rs#L6-L46

### `5ecdd62` test(token): cover fresh public recipient authorization

This commit does not change token behavior. It makes the new public-recipient contract explicit in docs and integration tests, so the suite now shows that the old raw public auto-create flow is no longer assumed, while a fresh public recipient still works if that recipient explicitly authorizes the transaction.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/token/core/src/lib.rs#L10-L16

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/integration_tests/tests/token.rs#L218-L479

https://github.com/logos-blockchain/logos-execution-zone/pull/414

### `b8d3d2f` fix(amm)!: require auth for fresh user lp holding

This is a breaking client-contract change. First-time pool creation still works in one transaction, but only if `user_lp` is authorized when the LP holding starts from the default state, which keeps the original UX alive under the new LEZ claim rules instead of forcing a silent pre-create step.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/amm/src/new_definition.rs#L76-L85

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/integration_tests/tests/amm.rs#L939-L1204

https://github.com/logos-blockchain/logos-execution-zone/pull/414

### `5c7e6bb` fix(ata)!: require recipient ata for transfers

This is a breaking ATA API change, but it preserves the higher-level ATA UX. Transfers to a fresh recipient ATA work again because ATA now receives the recipient owner, derives the recipient PDA itself, and authorizes the downstream token call correctly. The API is now ATA-to-ATA, not ATA-to-arbitrary-holding.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/ata/core/src/lib.rs#L21-L35

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/ata/src/transfer.rs#L7-L54

https://github.com/logos-blockchain/logos-execution-zone/pull/414

### `f6168fe` test(amm): assert pool pda claim on new definition

This is guardrail coverage for the new PDA claim contract introduced by LEZ. AMM tests now assert not only the pool data, but also the exact `Claim::Pda(...)` metadata that makes the pool claim valid.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/amm/src/tests.rs#L1927-L1936

https://github.com/logos-blockchain/logos-execution-zone/pull/414

### `7c34caa` test(ata): assert create chained call details

This is guardrail coverage for the ATA-to-token bridge contract. ATA create now proves the exact authorization and PDA seed it passes into token initialization, so a future refactor cannot silently break that handoff.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/ata/src/tests.rs#L75-L96

https://github.com/logos-blockchain/logos-execution-zone/pull/414

### `c989253` test(token): cover claim migration expectations

This commit finishes the token-side migration coverage and makes the new claim rules visible in unit tests. It covers both sides of the conditional behavior: default destinations require a claim, while already initialized destinations do not.

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/token/src/tests.rs#L577-L730

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/token/src/tests.rs#L853-L975

https://github.com/logos-blockchain/lez-programs/blob/90ca372de4ca829623712fa86c540924ed1b53cc/token/src/tests.rs#L1113-L1130

https://github.com/logos-blockchain/logos-execution-zone/pull/414

### `90ca372` style(tests): format updated coverage

This is a formatting-only follow-up after the new coverage landed. It has no behavior change of its own.

### `3d02766` fix(ata): rename recipient to recipient_owner and add recipient_ata account

Just a late missing IDL generation. 

### `24edac5` chore(docs/ata/create): clarify authorization check details for ATA creation

Just a small enhancement on code comments

### `cd58219` chore(token): update authorization requirements for token accounts in instructions and assertions

Added explicit checks for the `Claim` authorizations, so when they are missed a proper error would appear instead of a out of context error message. Also, added testes to verify the presence of these checks. 

## Validation

All of the following were run locally and passed:

```bash
taplo fmt --check .
```

### DIFF
--- a/token/src/initialize.rs
+++ b/token/src/initialize.rs
@@ -1,6 +1,6 @@
 use nssa_core::{
     account::{Account, AccountWithMetadata, Data},
-    program::AccountPostState,
+    program::{AccountPostState, Claim},
 };
 use token_core::{TokenDefinition, TokenHolding};
 
@@ -12,6 +12,10 @@ pub fn initialize_account(
         account_to_initialize.account,
         Account::default(),
         "Only Uninitialized accounts can be initialized"
+    );
+    assert!(
+        account_to_initialize.is_authorized,
+        "Account to initialize must be authorized"
     );
 
     // TODO: #212 We should check that this is an account owned by the token program.
@@ -29,6 +33,6 @@ pub fn initialize_account(
 
     vec![
         AccountPostState::new(definition_post),
-        AccountPostState::new_claimed(account_to_initialize),
+        AccountPostState::new_claimed(account_to_initialize, Claim::Authorized),
     ]
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -599,6 +599,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clock_core"
+version = "0.1.0"
+source = "git+https://github.com/logos-blockchain/logos-execution-zone.git?tag=v0.2.0-rc1#35d8df0d031315219f94d1546ceb862b0e5b208f"
+dependencies = [
+ "borsh",
+ "nssa_core",
+]
+
+[[package]]
 name = "cobs"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1821,10 +1830,11 @@ checksum = "a5b0c77c1b780822bc749a33e39aeb2c07584ab93332303babeabb645298a76e"
 [[package]]
 name = "nssa"
 version = "0.1.0"
-source = "git+https://github.com/logos-blockchain/logos-execution-zone.git?rev=ffcbc15972adbf557939bf3e2852af276422631b#ffcbc15972adbf557939bf3e2852af276422631b"
+source = "git+https://github.com/logos-blockchain/logos-execution-zone.git?tag=v0.2.0-rc1#35d8df0d031315219f94d1546ceb862b0e5b208f"
 dependencies = [
  "anyhow",
  "borsh",
+ "clock_core",
  "hex",
  "k256",
  "log",
@@ -1842,7 +1852,7 @@ dependencies = [
 [[package]]
 name = "nssa_core"
 version = "0.1.0"
-source = "git+https://github.com/logos-blockchain/logos-execution-zone.git?rev=ffcbc15972adbf557939bf3e2852af276422631b#ffcbc15972adbf557939bf3e2852af276422631b"
+source = "git+https://github.com/logos-blockchain/logos-execution-zone.git?tag=v0.2.0-rc1#35d8df0d031315219f94d1546ceb862b0e5b208f"
 dependencies = [
  "base58",
  "borsh",
@@ -2962,8 +2972,8 @@ dependencies = [
 
 [[package]]
 name = "spel-framework-core"
-version = "0.1.0"
-source = "git+https://github.com/logos-co/spel.git?rev=57201f64b4542bb3f592bc9a0aa654c47aa908a6#57201f64b4542bb3f592bc9a0aa654c47aa908a6"
+version = "0.2.0"
+source = "git+https://github.com/logos-co/spel.git?tag=v0.2.0-rc.2#9005e9fbbd78b0530412f9987273f753ed32eb2d"
 dependencies = [
  "borsh",
  "nssa_core",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,8 @@ exclude = [
 resolver = "2"
 
 [workspace.dependencies]
-nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", rev = "ffcbc15972adbf557939bf3e2852af276422631b", features = ["host"] }
-nssa = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", rev = "ffcbc15972adbf557939bf3e2852af276422631b", features = ["test-utils"] }
+nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "v0.2.0-rc1", features = ["host"] }
+nssa = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "v0.2.0-rc1", features = ["test-utils"] }
 token_core = { path = "token/core" }
 token_program = { path = "token" }
 amm_core = { path = "amm/core" }
@@ -33,4 +33,3 @@ borsh = { version = "1.0", features = ["derive"] }
 risc0-zkvm = { version = "=3.0.5" }
 serde_json = "1.0"
 tokio = { version = "1.28.2", features = ["net", "rt-multi-thread", "sync", "macros"] }
-
--- a/amm/Cargo.toml
+++ b/amm/Cargo.toml
@@ -4,6 +4,6 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", rev = "ffcbc15972adbf557939bf3e2852af276422631b", features = ["host"] }
+nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "v0.2.0-rc1", features = ["host"] }
 amm_core = { path = "core" }
 token_core = { path = "../token/core" }
--- a/amm/core/Cargo.toml
+++ b/amm/core/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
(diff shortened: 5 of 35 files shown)


token/src/initialize.rs
```diff
@@ -29,6 +29,6 @@ pub fn initialize_account(
 
     vec![
         AccountPostState::new(definition_post),
-        AccountPostState::new_claimed(account_to_initialize),
+        AccountPostState::new_claimed(account_to_initialize, Claim::Authorized),
```
0x-r4bbit, line 36: Actually here... if the account to be initialized happens to be an ATA, doesn't this need to be `Claim::PDA(seed)`?

3esmit:
# ATA Transfer UX Adaptation

`ATA::Transfer` moves tokens from the owner's ATA to a destination holding chosen by the caller, and the UX of ATA was defined to transfer to "any account; auto-created if default". 

https://github.com/logos-blockchain/lez-programs/blob/1f8eea84422d18efab1f3e271e292ddab2ee8b95/ata/core/src/lib.rs#L27

With the changes introduced by LEZ update, a default public account can no longer be materialized “for free”. It must be backed either by direct authorization for that account, or by a valid PDA proof for that account. Before the LEZ update, ATA transfer could get away with forwarding the recipient holding into token transfer and letting the old flow work. After the update, that stopped being valid for a fresh recipient ATA.

## How `associated_token_account` in https://github.com/logos-blockchain/logos-execution-zone/pull/414 solved this

LEZ did not solve this by keeping inline ATA creation inside `Transfer`.

Instead, LEZ solved it with a stricter 2 step flow for a new recipient:

- derive the recipient's canonical ATA
- create that ATA first
- then transfer to that existing ATA

The low-level `associated_token_account::Transfer` in LEZ still uses `(owner, sender_ata, recipient)` shape, so it don't have enough information to prove a fresh recipient ATA inline.

This solution was used because its the most straightforward, and after PR `#414`, a fresh public account could no longer be created during transfer unless the program could prove it was allowed to create that exact account, so it had to first create the ATA. 

TLDR how they solved:

- `Create` for a new recipient ATA
- `Transfer` only after that ATA already exists

This kept transfers working, but it changed the user flow for a new recipient from one step to two steps.

## ATA behavior before commit `5c7e6bb` - `fix(ata)!: require recipient ata for transfers`

The behavior that was implemented here strictly copied how it was implemented in PR #414

So, in practice, `b1a5afb` still behaved like "send to an existing recipient holding chosen by the caller." The integration test at that commit matched this: it pre-created the recipient ATA before calling `ATA::Transfer`.

## ATA behavior after `5c7e6bb` - `fix(ata)!: require recipient ata for transfers`

`ATA::Transfer` transfers to the recipient's canonical ATA, and that exact ATA may be initialized inline during the transfer if it is still default.

Therefore:

- The destination is not any token holding chosen by the caller.
- The destination must be the standard ATA that belongs to the recipient for that token.
- If that ATA already exists, the transfer uses it.
- If that ATA does not exist yet, the same transfer can initialize that exact ATA and send the tokens into it.

This means the current design is narrower than the old "send to any recipient holding" model, but it still keeps the one-step user flow for sending to a recipient who does not yet have an ATA. 

### Alternative: Don't change `Transfer` but add an helper function

Remove the "any account; auto-created if default" from `Transfer` requirements. 
- Explicitaly say that new recipient must use `Create` and than `Transfer`.
- Revert changes in `Transfer`, which require the recipient ATA to already exist.
- Add a separate `CreateAndTransfer` instruction to replicate the UX 1 call of Transfer on a fresh account, using the current commited code in `Transfer`.

token/src/initialize.rs
3esmit: I followed the implementation in [LEZ PR 414 diff](https://github.com/logos-blockchain/logos-execution-zone/pull/414/changes#diff-baa489675974e9306ac1606bc10985371a7c2364e947747785747b68b874ae71).

PR #65, opened 2026-04-22
feat(amm): route user deposits and LP burns through ATA program

### LINKED ISSUE #11: Add integration of ATAs in AMM
This needs to be researched how exactly the ATA program fits into the AMM program.
Explore possible options, decide which one makes most sense, implement it.

0x-r4bbit:
Add `owner` and `ata_program_id` parameters to `add_liquidity`, `remove_liquidity`, `swap_exact_input`, and `swap_exact_output`. User deposit-side transfers now emit `ATA::Transfer` chained calls instead of `Token::Transfer` directly, and LP burns emit `ATA::Burn` instead of `Token::Burn`. Vault withdrawal chained calls are unchanged.

- Add `ata_program_id` field to `AddLiquidity`, `RemoveLiquidity`, `SwapExactInput`, and `SwapExactOutput` instruction variants in `amm_core`
- Add `ata_core` dependency to `amm_program` and guest crates
- Update guest binary, unit tests, and integration tests to supply the new `owner` account and `ata_program_id` at every call site
- Regenerate `artifacts/amm-idl.json`

Closes #11

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,6 +54,7 @@ name = "amm_program"
 version = "0.1.0"
 dependencies = [
  "amm_core",
+ "ata_core",
  "nssa_core",
  "token_core",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,12 +10,14 @@ members = [
     "ata",
     "ata/methods",
     "integration_tests",
+    "integration_tests/malicious_ata_methods",
     "tools/idl-gen",
 ]
 exclude = [
     "token/methods/guest",
     "amm/methods/guest",
     "ata/methods/guest",
+    "integration_tests/malicious_ata_methods/guest",
 ]
 resolver = "2"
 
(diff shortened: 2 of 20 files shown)


3esmit:
Im reviewing this code, I am trying to exploit the lack of trustful relationship between amm and ata transfer. Dont merge it before my final review

0x-r4bbit:
Thank you @3esmit for laying out the vulnerabilities.
While we've already (soft)decided that we won't integrate ATA into the AMM, the issue you're demonstrating here can be generalized and applied to other programs as well, which I think is a great learning.

Might be worth looking into patterns on how to ensure a program has a program dependency it can trust (maybe via a config account for the program or something like that).

3esmit (changes requested):
This can't be merged because it allows an untrusted program to steal from AMM, as AMM trusts the program is doing the actual transfers but can't actually checks they happened due how chainedcalls are designed to be proof compatible. 
If this ever would have to be approved, either:
- Force a requirement on verified programids
- Change the chainedcalls to a 2 phase action where AMM verify that transfers acutally occured.

0x-r4bbit:
Closing this, as we're not moving forward with integrating ATA into AMM.

PR #67, opened 2026-04-23
feat(amm)!: add transaction deadlines to swap and liquidity instructions

### LINKED ISSUE #8: Implement transaction expiry in AMM program
To prevent reply attacks, transactions should have a `deadline` so they expire after a certain time.

0x-r4bbit:
AddLiquidity, RemoveLiquidity, SwapExactInput, and SwapExactOutput now require a `deadline: u64` field (Unix timestamp in milliseconds). The guest program reads the on-chain CLOCK_01 account and panics if `clock.timestamp > deadline`, preventing stale transactions from being executed. NewDefinition is excluded since it sets the initial price explicitly with fixed amounts — there is no time-sensitive price risk.

BREAKING CHANGE: AddLiquidity, RemoveLiquidity, SwapExactInput, and SwapExactOutput instruction variants now require a `deadline` field, and `CLOCK_01_PROGRAM_ACCOUNT_ID` must be included as an account in each of these transactions. The clock account is populated by the LEZ clock program — a dedicated system transaction that writes the current timestamp into the clock accounts. The AMM guest reads the account data directly; there is no implicit timestamp injection from the runtime.

Closes #8

### DIFF
--- a/amm/core/src/lib.rs
+++ b/amm/core/src/lib.rs
@@ -35,6 +35,8 @@ pub enum Instruction {
         token_b_amount: u128,
         fees: u128,
         amm_program_id: ProgramId,
+        /// Unix timestamp (milliseconds) after which this transaction is invalid.
+        deadline: u64,
     },
 
     /// Adds liquidity to the Pool
@@ -51,6 +53,8 @@ pub enum Instruction {
         min_amount_liquidity: u128,
         max_amount_to_add_token_a: u128,
         max_amount_to_add_token_b: u128,
+        /// Unix timestamp (milliseconds) after which this transaction is invalid.
+        deadline: u64,
     },
 
     /// Removes liquidity from the Pool
@@ -67,6 +71,8 @@ pub enum Instruction {
         remove_liquidity_amount: u128,
         min_amount_to_remove_token_a: u128,
         min_amount_to_remove_token_b: u128,
+        /// Unix timestamp (milliseconds) after which this transaction is invalid.
+        deadline: u64,
     },
 
     /// Swap some quantity of Tokens (either Token A or Token B)
@@ -77,12 +83,13 @@ pub enum Instruction {
     /// - Vault Holding Account for Token A (initialized)
     /// - Vault Holding Account for Token B (initialized)
     /// - User Holding Account for Token A
-    /// - User Holding Account for Token B Either User Holding Account for Token A or Token B is
-    ///   authorized.
+    /// - User Holding Account for Token B; either is authorized.
     SwapExactInput {
         swap_amount_in: u128,
         min_amount_out: u128,
         token_definition_id_in: AccountId,
+        /// Unix timestamp (milliseconds) after which this transaction is invalid.
+        deadline: u64,
     },
 
     /// Swap tokens specifying the exact desired output amount,
@@ -93,12 +100,13 @@ pub enum Instruction {
     /// - Vault Holding Account for Token A (initialized)
     /// - Vault Holding Account for Token B (initialized)
     /// - User Holding Account for Token A
-    /// - User Holding Account for Token B Either User Holding Account for Token A or Token B is
-    ///   authorized.
+    /// - User Holding Account for Token B; either is authorized.
     SwapExactOutput {
         exact_amount_out: u128,
         max_amount_in: u128,
         token_definition_id_in: AccountId,
+        /// Unix timestamp (milliseconds) after which this transaction is invalid.
+        deadline: u64,
     },
 
     /// Sync pool reserves with current vault balances.
--- a/amm/methods/guest/Cargo.toml
+++ b/amm/methods/guest/Cargo.toml
@@ -10,7 +10,7 @@ name = "amm"
 path = "src/bin/amm.rs"
 
 [dependencies]
-spel-framework = { git = "https://github.com/logos-co/spel.git", tag = "v0.2.0-rc.2", package = "spel-framework" }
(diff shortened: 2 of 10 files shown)


integration_tests/Cargo.toml
```diff
@@ -9,6 +9,7 @@ nssa_core = { workspace = true, features = ["host"] }
 amm_core = { workspace = true }
 token_core = { workspace = true }
 ata_core = { workspace = true }
+clock_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "v0.2.0-rc1" }
```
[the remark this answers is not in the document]
0x-r4bbit: Done

gravityblast (approved):

amm/methods/guest/src/bin/amm.rs
3esmit: This PR adds deadline enforcement to `AddLiquidity`, `RemoveLiquidity`, `SwapExactInput`, and `SwapExactOutput`, but `NewDefinition` still returns an output with no timestamp validity window. If the goal is to close AMM transaction expiry generally, can we add the same `deadline` field and `.with_timestamp_validity_window(..deadline)` here, or explicitly leave pool initialization for a separate issue/PR?

amm/core/src/lib.rs
3esmit, line 33: This changes keep `NewDefinition` without a `deadline`, and `amm/methods/guest/src/bin/amm.rs:21` to `amm/methods/guest/src/bin/amm.rs:49` still returns `SpelOutput::with_chained_calls(...)` without a timestamp validity window. That path consumes the user's token A and token B holdings and initializes the pool at a signed price. A signed transaction can therefore remain executable indefinitely as long as the relevant nonces still match, even though issue #8 asks for transaction expiry in the AMM program to prevent stale execution.

3esmit (changes requested):
Add deadline on the pool initialization, filling the `NewDefinition` expiry gap. This PR adds deadline enforcement to `AddLiquidity`, `RemoveLiquidity`, `SwapExactInput`, and `SwapExactOutput`, but pool initialization can still execute later with no time boundary if the signed holding nonces remain unchanged. The concrete risk is stale initial pricing: a user can sign pool creation at one token ratio, the transaction can sit unsubmitted, and later execution can initialize the pool at that old ratio, leaving the first LP position exposed to immediate arbitrage. Please either add the same deadline/timestamp validity window to `NewDefinition`, or explicitly split pool-initialization expiry into a follow-up and avoid presenting this as complete AMM transaction-expiry coverage.

amm/methods/guest/Cargo.toml
```diff
@@ -10,7 +10,7 @@ name = "amm"
 path = "src/bin/amm.rs"
 
 [dependencies]
```
3esmit, line 12: Wouldn't be better to make a release on the spel and tag that to the release version?

integration_tests/tests/amm.rs
```diff
@@ -1671,6 +1680,119 @@ fn amm_fee_accumulates_across_multiple_swaps_and_pays_out_on_remove() {
     );
 }
 
+#[test]
+fn amm_swap_rejects_expired_deadline() {
```
3esmit, line 1685: The new expired-deadline coverage starts here, but it only covers `SwapExactInput`, `AddLiquidity`, and `RemoveLiquidity`. Can we add the same expired-deadline test for `SwapExactOutput` and match `Err(NssaError::OutOfValidityWindow)` instead of only checking `is_err()`?

3esmit (changes requested):
Improve test coverage. The implementation adds a deadline to `SwapExactOutput`, but the new expired-deadline integration tests only cover exact-input swap, add liquidity, and remove liquidity. Please add an expired-deadline `SwapExactOutput` test and assert `NssaError::OutOfValidityWindow` so the tests prove the deadline mechanism, not just any failure.

3esmit:
> `CLOCK_01_PROGRAM_ACCOUNT_ID` must be included as an account in each of these transactions. The clock account is populated by the LEZ clock program — a dedicated system transaction that writes the current timestamp into the clock accounts. The AMM guest reads the account data directly; there is no implicit timestamp injection from the runtime.


The current PR body still describes a CLOCK_01 account and direct clock-account reads, but the diff now documents and implements timestamp validity windows on the program output with no clock account. Update the PR body before merge so integrators do not add a nonexistent account to the AMM instruction interface.

amm/methods/guest/src/bin/amm.rs
0x-r4bbit: I was considering it but then dropped the deadline for this function.
I don't think there's a downside to having the initialization happen later than some deadline.
The outcome would be the same.

This is not necessarily true for swapping and adding liquidity.

0x-r4bbit:
@3esmit address all of yoru comments

integration_tests/tests/amm.rs
0x-r4bbit: Done

amm/methods/guest/Cargo.toml
0x-r4bbit: We need to wait for a release (the release will also just point at a commit so we're not loosing anything right now)

amm/core/src/lib.rs
0x-r4bbit: Updated.

amm/methods/guest/src/bin/amm.rs
0x-r4bbit: Okay, added it anyways now.

3esmit: I think is a good feature to have, and if everything works as intended (pool gets initialized shortly after user intent), the existence of this deadline would be invisible, and is kept as a protection layer.

3esmit (approved):

PR #66, opened 2026-04-23
`feat(amm-ui): add liquidity pool management flow`

3esmit:
## Summary

Adds complete AMM liquidity UI covering all five issues: pool position summary, add-liquidity preview, remove-liquidity preview, slippage/min-received controls, and mock confirmation flow.

## Changes

- Replaces placeholder `amm-ui/Main.qml` with `amm-ui/qml/Main.qml`.
- Updates `amm-ui/metadata.json` to load `qml/Main.qml`.
- Adds `amm-ui/qml/pages/LiquidityPage.qml` as main page owner.
- Adds reusable QML components under `amm-ui/qml/components/`.
- Adds `amm-ui/qml/state/DummyPoolState.qml` for local preview state.
- Keeps scope UI-only: no program logic, transaction path, or IDL changes.

## Structure choices

- `qml/` becomes clear QML import/package root. Common QML layout; also simpler for Nix packaging because metadata points to one stable entry.
- `Main.qml` stays thin. Page state and layout move to `LiquidityPage.qml`, so app shell does not become feature code.
- Components split by responsibility: token inputs, tabs, summary rows, slippage control, confirmation dialog, toast. This keeps each issue reviewable and future backend wiring easier.
- Dummy state stays isolated because issues require preview/mock behavior, not real transactions.

### `e5c93c2` - `feat(amm/ui): add pool position summary (#60)`

Fixes #60.

Adds pool position summary UI with token pair, fee tier, user LP balance, pool share, owned token estimates, pool reserves, and LP supply. Establishes page structure and reusable summary rows used by later commits.

### `88e3f40` - `feat(amm/ui): add liquidity deposit preview (#61)`

Fixes #61.

Adds add-liquidity form with token amount inputs, balances, MAX action, pool-ratio helper, estimated LP tokens, and input validation. Preview updates before submit so user can inspect deposit effect.

### `e92a94d` - `feat(amm/ui): add remove-liquidity preview (#62)`

Fixes #62.

Adds remove-liquidity form with LP burn input, 25/50/75/MAX presets, share slider, expected token withdrawal, and post-withdrawal position preview. Uses same local pool state model for consistent add/remove math.

### `a65621c` - `feat(amm/ui): add slippage min received controls (#63)`

Fixes #63.

Adds slippage presets, custom tolerance input, min-received rows, clamping, and warning states. Shared slippage component keeps add/remove behavior consistent and avoids duplicated tolerance logic.

[...]

PR #72, opened 2026-04-28
feat(amm-ui): add navbar to switch between views

0x-r4bbit:


0x-r4bbit:
This has landed with #66 

Sorry for messing this up and you couldn't review first..

PR #78, opened 2026-05-04
chore(amm): add defensive check for lp token solvency

0x-r4bbit:
This check is added to fulfill the program invariant that no more tokens than owned can be burned. This was not a bug before, because the `token` program will revert on `Transfer::Burn` when one tries to burn more tokens than available.

So this change is merely for making the invariant explicit.

### DIFF
--- a/amm/src/remove.rs
+++ b/amm/src/remove.rs
@@ -101,6 +101,10 @@ pub fn remove_liquidity(
         pool_def_data.liquidity_pool_supply > MINIMUM_LIQUIDITY,
         "Pool only contains locked liquidity"
     );
+    assert!(
+        remove_liquidity_amount <= user_lp_balance,
+        "Remove amount exceeds user LP balance"
+    );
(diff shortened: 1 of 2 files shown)


amm/src/remove.rs
```diff
@@ -91,6 +91,10 @@ pub fn remove_liquidity(
         pool_def_data.liquidity_pool_supply > MINIMUM_LIQUIDITY,
         "Pool only contains locked liquidity"
     );
+    assert!(
```
3esmit, line 104: The guard is the right place for this invariant. Can we add a direct `amm_program` unit test that expects `Remove amount exceeds user LP balance`? The existing integration test only asserts generic failure and would still pass if this explicit AMM-side check regressed back to relying on the downstream token burn.

3esmit (commented):
This PR is fine, but I suggest adding test to ensure this won't be accidentally removed.

amm/src/remove.rs
0x-r4bbit: Done!

PR #84, opened 2026-05-04
fix(ata): validate token dispatch and existing occupants

3esmit:
Closes #83

## Summary

This PR makes ATA token operations dispatch only to the canonical token program and tightens idempotent `Create` handling for already occupied ATA PDAs.    

ATA now receives the trusted token program ID from guest wiring instead of deriving downstream token dispatch from caller-supplied account metadata. Existing ATA occupants are accepted as idempotent only when they are canonical token holdings for the requested token definition.

## Changes

- Threaded the trusted token program ID through ATA create, transfer, and burn logic.
- Wired ATA guest dispatch to `token_methods::TOKEN_ID` without adding a caller-supplied instruction argument or changing the ATA IDL.
- Added canonical ownership and decode checks for token definitions and ATA holdings before emitting chained token calls.
- Made idempotent `Create` reject malformed, foreign-owned, or wrong-definition ATA PDA occupants.
- Made `Burn` reject foreign-owned token definitions and holder/definition mismatches before forwarding the burn.
- Updated ATA core comments so they no longer describe token dispatch as account-owner-derived.
- Added unit and integration coverage for non-canonical token owners, malformed existing ATA occupants, and burn definition mismatches.

## Validation

The following checks passed:

```bash
taplo fmt --check .
cargo +1.94.0 run -p idl-gen -- ata/methods/guest/src/bin/ata.rs > /tmp/lez-ata-idl.json
diff -u artifacts/ata-idl.json /tmp/lez-ata-idl.json
```


3esmit:
This should not be merged yet. I am researching on a proper solution (actually a proposal/feature request for LEZ), which would not require hardcoding programid, because this limitation causes a lot of problems down the road for the future of this ecossystem.

PR #85, opened 2026-05-05
feat: make use of spel's `[#account_type]` directive

### LINKED ISSUE #49: Make use of SPELs new `#[account_type]` directive
This has been introduced in https://github.com/logos-co/spel/commit/62f91e22777689e4a1928937513bb1fa9177377a

We can apply these to our programs and update the IDL

0x-r4bbit:
This annotates custom account data as `[#account_type]`, to allow for deserializing abritrary LEZ account data.

Closes #49

### DIFF
--- a/amm/core/src/lib.rs
+++ b/amm/core/src/lib.rs
@@ -6,6 +6,7 @@ use nssa_core::{
     program::{PdaSeed, ProgramId},
 };
 use serde::{Deserialize, Serialize};
+use spel_framework_macros::account_type;
 
 // These stable seed bytes are part of the PDA derivation scheme and must stay unchanged for
 // compatibility.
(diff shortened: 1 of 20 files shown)


gravityblast (approved):

amm/core/src/lib.rs
```diff
@@ -252,7 +254,7 @@ pub fn compute_liquidity_token_pda_seed(pool_id: AccountId) -> PdaSeed {
 }
 
 pub fn compute_lp_lock_holding_pda(amm_program_id: ProgramId, pool_id: AccountId) -> AccountId {
-    AccountId::from((&amm_program_id, &compute_lp_lock_holding_pda_seed(pool_id)))
+    AccountId::for_public_pda(&amm_program_id, &compute_lp_lock_holding_pda_seed(pool_id))
```
3esmit, line 265: Confirm this is the intended change. What this actually does?

3esmit (approved):

amm/core/src/lib.rs
0x-r4bbit: This is a breaking change that was introduced in LEZ v0.2.0-rc3. It introduces public and private PDAs, and these are dedicated functions to derive them.

I've extracted this change out into a separate PR that updates the LEZ dependency and makes the necessary changes according to the breaking changes introduced by that version.

PR #103, opened 2026-05-11
fix(ata): lock down `ATA::Transfer` recipient contract

### LINKED ISSUE #53: Clarify ATA transfer recipient contract after the LEZ update
## Summary

The LEZ update changed the ATA transfer surface on `main`.

`ATA::Transfer` now accepts a generic recipient token holding account instead of encoding extra recipient-side ATA structure in the instruction surface. That may be the right direction, but the current contract still needs to be made explicit for integrators and locked down in tests.

The core ATA instruction docs now define `Transfer` as a 3-account flow whose destination is a generic recipient token holding.

https://github.com/logos-blockchain/lez-programs/blob/471abef7192b4a77f341eab8bc49c7c4c79e48fd/ata/core/src/lib.rs#L21-L33

The guest wrapper then narrows that description by saying the recipient holding must already be initialized.

https://github.com/logos-blockchain/lez-programs/blob/471abef7192b4a77f341eab8bc49c7c4c79e48fd/ata/methods/guest/src/bin/ata.rs#L31-L49

That overall contract is still not crisp enough, because the ATA implementation delegates recipient behavior to the token layer, where default-account handling is still influenced by LEZ claim semantics.

## Problem

After the LEZ update:

- `ata_core::Instruction::Transfer` documents a 3-account flow:
  - authorized owner
  - sender ATA
  - recipient token holding
- the guest wrapper repeats that the recipient holding must already be initialized
- `ata_program::transfer` verifies the sender ATA and delegates the recipient behavior to the downstream token transfer
- `token::Transfer` still has default-recipient claim behavior through `Claim::Authorized`

The current ATA transfer implementation verifies the sender ATA, marks only that sender as authorized for the chained call, and forwards the recipient account as-is into `token::Transfer`.

https://github.com/logos-blockchain/lez-programs/blob/471abef7192b4a77f341eab8bc49c7c4c79e48fd/ata/src/transfer.rs#L7-L38

The downstream token transfer still treats a default recipient as materializable by cloning a zeroized holding shape from the sender.

https://github.com/logos-blockchain/lez-programs/blob/471abef7192b4a77f341eab8bc49c7c4c79e48fd/token/src/transfer.rs#L17-L21

That same token path still emits `Claim::Authorized` for a default recipient post-state.

https://github.com/logos-blockchain/lez-programs/blob/471abef7192b4a77f341eab8bc49c7c4c79e48fd/token/src/transfer.rs#L106-L109

This leaves an open contract question for the ATA layer:

- is the intended rule truly "recipient must already be initialized"
- or is the intended rule "recipient may be initialized, or may be default if the downstream authorization model allows it"

The current implementation may be acceptable either way, but the intended rule should be stated and tested directly.

## Why It Matters

- ATA is now exposing a broader recipient model after the LEZ update, so client code needs a precise behavioral contract.
- If the rule is "initialized recipient only", callers should get a clear ATA-level failure model instead of having to infer behavior from token/runtime claim handling.
- If broader recipient cases are intentionally supported, that support should be documented and tested explicitly.
- Without sharper coverage, future refactors could change recipient behavior without making that change obvious in review.

The current ATA integration coverage only locks down the initialized-recipient happy path.

https://github.com/logos-blockchain/lez-programs/blob/471abef7192b4a77f341eab8bc49c7c4c79e48fd/integration_tests/tests/ata.rs#L215-L261

That means the repository still does not make the intended behavior obvious for default recipients, malformed token holdings, or other recipient-shape boundary cases.

## Proposed Scope

- Define the intended recipient contract for the current post-LEZ `ATA::Transfer` API.
- Align the core docs, guest wrapper docs, and IDL with that contract.
- Decide whether ATA should:
  - enforce initialized-recipient requirements directly
  - or explicitly rely on downstream token/runtime claim semantics for some recipient shapes
- Add boundary coverage for the recipient cases the project wants to support or reject.

## Acceptance Criteria

- The repository documents one clear recipient contract for `ATA::Transfer` on the current LEZ-based surface.
- Core instruction comments, guest wrapper comments, and emitted IDL all describe the same contract.
- Tests cover the intended supported path for an initialized recipient holding.
- Tests also cover the intended behavior for unsupported or special recipient states, such as:
  - default recipient account
  - malformed recipient token data
  - recipient with mismatched definition
- The final behavior does not require integrators to reverse-engineer ATA semantics from token internals.

## Validation Hints

- Run ATA unit tests and ATA integration tests after tightening the contract.
- Add explicit negative tests for recipient-shape failures rather than relying only on the happy path.
- Verify that any instruction docs or generated artifacts exposed to clients match the final behavior.

## Notes

- This issue is about clarifying and locking the ATA semantics that exist after the LEZ update.
- It should not assume that the older ATA transfer shape or older recipient expectations need to be restored.

3esmit:
Closes #53

## Summary

After the LEZ update, `ATA::Transfer` accepts a generic recipient token holding account, but the existing contract was only documented as "must be initialized" and not enforced at the ATA layer. The downstream `token::Transfer` still has default-recipient claim handling via `Claim::Authorized`, so callers had to reverse-engineer the boundary from token/runtime semantics.

This PR pins the recipient contract at the ATA layer and locks it down with explicit tests.

## Decision: enforce "initialized recipient" at the ATA layer

`ata_program::transfer` now requires the recipient holding to be:

- already initialized (not `Account::default()`),
- owned by the same token program as the sender ATA,
- decode to a valid `TokenHolding`,
- and point at the same token definition as the sender.

Each boundary fails with a distinct ATA-level assertion message rather than being silently materialized by the downstream `token::Transfer` (e.g. via the default-recipient `Claim::Authorized` path) or surfacing as a "Mismatch Token Definition" panic from the token layer.

The downstream `token::Transfer` default-recipient logic is left in place — it remains usable from the direct token program surface — but it is no longer reachable via the ATA transfer path, which is the surface this issue is about.

## Changes

- `ata/src/transfer.rs`: enforce the recipient contract on the ATA-side before constructing the chained `Token::Transfer` call.
- `ata/core/src/lib.rs`: expand the `Instruction::Transfer` doc-comment to describe all three recipient invariants (initialized, same token program, same definition).
- `ata/methods/guest/src/bin/ata.rs`: align the guest wrapper doc-comment with the core contract.
- `ata/src/tests.rs`: add 5 unit tests covering the contract:
  - happy path emits the expected chained `Token::Transfer`,
  - missing owner authorization,
  - default recipient,
  - foreign-owned recipient,
  - malformed recipient data,
  - mismatched definition recipient.
- `integration_tests/tests/ata.rs`: add 2 integration tests covering the end-to-end runtime rejection of:
  - default recipient,
  - mismatched definition recipient.

The wire IDL (`artifacts/ata-idl.json`) is unchanged — this PR clarifies and enforces the existing instruction surface rather than reshaping it.

## Why not the broader contract

The other option was to keep ATA delegating to the downstream token layer and let the default-recipient `Claim::Authorized` path materialize a fresh recipient holding when applicable. That would have required the ATA-level docs to describe a multi-modal recipient behavior driven by token/runtime claim semantics, which is exactly the reverse-engineering hazard #53 flags. The single "initialized recipient only" rule is narrower and makes the failure model explicit at the surface clients interact with.

## Validation

```bash
taplo fmt --check .
RISC0_DEV_MODE=1 cargo +1.94.0 test -p ata_program
RISC0_DEV_MODE=1 cargo +1.94.0 test -p integration_tests --test ata
cargo +1.94.0 run -q -p idl-gen -- ata/methods/guest/src/bin/ata.rs > /tmp/lez-ata-idl.json
diff -u artifacts/ata-idl.json /tmp/lez-ata-idl.json
```

All commands pass; the ATA IDL is unchanged.

### DIFF
--- a/ata/src/transfer.rs
+++ b/ata/src/transfer.rs
@@ -1,5 +1,5 @@
 use nssa_core::{
-    account::AccountWithMetadata,
+    account::{Account, AccountWithMetadata},
     program::{AccountPostState, ChainedCall, ProgramId},
 };
 use token_core::TokenHolding;
@@ -13,11 +13,38 @@ pub fn transfer_from_associated_token_account(
 ) -> (Vec<AccountPostState>, Vec<ChainedCall>) {
     let token_program_id = sender_ata.account.program_owner;
     assert!(owner.is_authorized, "Owner authorization is missing");
-    let definition_id = TokenHolding::try_from(&sender_ata.account.data)
+    let sender_definition_id = TokenHolding::try_from(&sender_ata.account.data)
         .expect("Sender ATA must hold a valid token")
         .definition_id();
-    let sender_seed =
-        ata_core::verify_ata_and_get_seed(&sender_ata, &owner, definition_id, ata_program_id);
+    let sender_seed = ata_core::verify_ata_and_get_seed(
+        &sender_ata,
+        &owner,
+        sender_definition_id,
+        ata_program_id,
+    );
+
+    // The recipient contract: ATA::Transfer requires a recipient token holding that is already
+    // initialized, owned by the same token program as the sender ATA, and that points at the same
+    // token definition as the sender. Anything else fails here rather than being silently
+    // materialized by the downstream token transfer (e.g. via `Claim::Authorized` on a default
+    // recipient), so integrators get an ATA-level failure rather than having to reverse-engineer
+    // token/runtime semantics.
+    assert_ne!(
+        recipient.account,
+        Account::default(),
+        "Recipient token holding must be initialized"
+    );
+    assert_eq!(
+        recipient.account.program_owner, token_program_id,
+        "Recipient must be owned by the same token program as the sender ATA"
+    );
+    let recipient_definition_id = TokenHolding::try_from(&recipient.account.data)
+        .expect("Recipient must hold a valid token")
+        .definition_id();
+    assert_eq!(
+        recipient_definition_id, sender_definition_id,
+        "Recipient and sender token definitions do not match"
+    );
 
     let post_states = vec![
         AccountPostState::new(owner.account.clone()),
--- a/ata/core/src/lib.rs
+++ b/ata/core/src/lib.rs
@@ -24,7 +24,10 @@ pub enum Instruction {
     /// Required accounts (3):
     /// - Owner account (authorized)
     /// - Sender ATA (owner's token holding)
-    /// - Recipient token holding (must be initialized)
+    /// - Recipient token holding. Must be:
+    ///   - already initialized (not a default account),
+    ///   - owned by the same token program as the sender ATA,
+    ///   - and point at the same token definition as the sender.
     ///
     /// `token_program_id` is derived from `sender_ata.account.program_owner`.
     Transfer {
--- a/ata/methods/guest/src/bin/ata.rs
+++ b/ata/methods/guest/src/bin/ata.rs
@@ -29,7 +29,8 @@ mod ata {
     }
 
     /// Transfer tokens FROM owner's ATA to a recipient token holding account.
-    /// The recipient holding account must already be initialized.
+    /// The recipient holding must already be initialized, be owned by the same token program
+    /// as the sender ATA, and point at the same token definition as the sender.
     #[instruction]
     pub fn transfer(
         owner: AccountWithMetadata,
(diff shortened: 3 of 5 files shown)


ata/src/transfer.rs
```diff
@@ -13,11 +13,38 @@ pub fn transfer_from_associated_token_account(
...
+    // initialized, owned by the same token program as the sender ATA, and that points at the same
+    // token definition as the sender. Anything else fails here rather than being silently
+    // materialized by the downstream token transfer (e.g. via `Claim::Authorized` on a default
+    // recipient), so integrators get an ATA-level failure rather than having to reverse-engineer
+    // token/runtime semantics.
+    assert_ne!(
+        recipient.account,
+        Account::default(),
+        "Recipient token holding must be initialized"
+    );
```
0x-r4bbit, line 36: I think this one here is not necessary. If the account isn't initialized, token problem would try to claim and fail because of no signature being available.

PR #104, opened 2026-05-11
feat(stablecoin): implement `open_position`

3esmit:
Closes #89

## Summary

Adds the `open_position` instruction to the Stablecoin Program. A caller can now lock
collateral into a per-owner [`Position`] account whose collateral is custodied by a vault
token holding initialized and transferred into in the same transaction.

## What Changed

- `stablecoin/core`
  - New `Position` struct (`collateral_vault_id`, `collateral_definition_id`,
    `collateral_amount`, `debt_amount`) with Borsh codec to/from `Data`.
  - New `OpenPosition { stablecoin_program_id, collateral_amount }` instruction variant.
  - New PDA helpers: `compute_position_pda{,_seed}` (seed derived from the owner address
    with a domain-separation tag) and `compute_position_vault_pda{,_seed}` (seed derived
    from the position id with its own tag so position and vault PDAs cannot collide).
  - `verify_position_and_get_seed` / `verify_position_vault_and_get_seed` mirror the ATA
    pattern: assert the supplied account id matches the derivation and return the seed for
    chained-call use.
  - Pulled `risc0-zkvm` (default-features off) into `stablecoin_core` to access the SHA
    impl used by PDA derivation, matching `amm_core` and `ata_core`.
- `stablecoin` (program crate)
  - New `open_position::open_position` host function:
    - Asserts owner and user-holding authorization, position and vault uninitialized.
    - Verifies the supplied position and vault account ids match the PDA derivations.
    - Cross-checks the supplied token definition against the user holding's
      `definition_id` and the user holding's `program_owner` so the chained
      `InitializeAccount` cannot be steered to an unrelated Token Program.
    - Returns five post-states (owner, position, vault, user holding, token definition).
      The position post-state is `Claim::Pda(position_seed)` with `program_owner` set to
      `stablecoin_program_id` and `data` encoding `Position { collateral_amount,
      debt_amount: 0, ... }`.
    - Emits two chained calls under the vault's PDA authority: `Token::InitializeAccount`
      and `Token::Transfer` (with the post-init vault as the recipient pre-state, mirroring
      `amm::new_definition`'s expected-pre-state pattern).
  - Added `token_core` as a direct dependency of `stablecoin_program`.
- `stablecoin/methods/guest`
  - New `open_position` `#[instruction]` wrapper threaded through the SPEL framework.
  - Added `token_core` dependency to the guest crate.
- Regenerated `artifacts/stablecoin-idl.json`.
- Tests: 11 new unit tests covering the happy path, all assertion paths, and PDA
  determinism / non-collision invariants.

## Justification

Issue #89 requires `open_position` to claim the position PDA from the owner's address,
initialize a vault token holding via a chained call with no user signature, transfer the
user's collateral into the vault, and persist `Position { collateral_amount, debt_amount:
0 }`. The implementation follows the existing ATA `create` and AMM `new_definition`
patterns for PDA-claimed accounts and same-transaction chained-call composition so the
new flow stays consistent with the rest of the program suite.

The `debt_amount` parameter is deliberately omitted, per the issue's note that it is
deferred to `generate_debt` (price feed + stability fees). `Position` already carries the
field so the future instruction can mutate it without a state-shape migration.

## Risks

- `Position` now stores `collateral_vault_id` and `collateral_definition_id` in addition
  to the two fields named in the issue. Both are re-derivable, so the redundancy costs a
  small amount of data per position, but keeps subsequent instructions (`deposit_collateral`,
  `withdraw_collateral`, `generate_debt`, ...) from re-deriving and re-validating the vault
  and collateral identity on every call. If this is undesirable they can be removed before
  any of those instructions ship.
- The vault PDA seed uses a domain tag distinct from the position seed
  (`stablecoin::position::vault::v1`). Changing either tag after merge would break
  on-chain address compatibility for any deployed position; the tags are constants in
  `stablecoin_core` to make this contract explicit.

## Validation

```bash
taplo fmt --check .
RISC0_SKIP_BUILD=1 cargo +1.94.0 test --workspace --exclude integration_tests
```

All four passed locally; the new `stablecoin_program` test module reports 12 passed.
Integration tests under `integration_tests/` need real RISC0 guest builds and are gated
on the `integration-tests` CI job rather than the local skip-build flow.


stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+        Self::try_from(data).expect("Position encoded data should fit into Data")
+    }
+}
+
+/// PDA seed for the [`Position`] account owned by `owner_id`.
+///
+/// Derived from the owner's address with a domain-separation tag so the resulting seed
+/// cannot collide with other stablecoin-managed PDAs that hash the same owner id.
+pub fn compute_position_pda_seed(owner_id: AccountId) -> PdaSeed {
+    use risc0_zkvm::sha::{Impl, Sha256};
```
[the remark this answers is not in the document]
3esmit: @Copilot 
False positive. `Sha256` is a trait, and `Impl::hash_bytes` is one of its
methods — Rust requires the trait to be in scope to call it via the impl
type. Dropping `Sha256` from the `use` fails the build with:

```
error[E0599]: no function or associated item named `hash_bytes` found
              for struct `risc0_zkvm::sha::Impl` in the current scope
  --> stablecoin/core/src/lib.rs:95:31
   |
95 |     out.copy_from_slice(Impl::hash_bytes(&bytes).as_bytes());
   |                               ^^^^^^^^^^ function or associated item
   |                                          not found in `risc0_zkvm::sha::Impl`
   |
   = help: items from traits can only be used if the trait is in scope
help: trait `Sha256` which provides `hash_bytes` is implemented but not
      in scope; perhaps you want to import it
   |
 3 + use risc0_zkvm::sha::Sha256;
   |
```

The identical `use risc0_zkvm::sha::{Impl, Sha256};` pattern is used in
`amm/core/src/lib.rs` (four call sites) and `ata/core/src/lib.rs` for the
same reason. No change here.

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+}
+
+/// PDA seed for the collateral vault token holding bound to a [`Position`].
+///
+/// Derived from the position's address with a distinct domain-separation tag so the vault
+/// id cannot collide with the position id even though both PDAs share the same program.
+pub fn compute_position_vault_pda_seed(position_id: AccountId) -> PdaSeed {
+    use risc0_zkvm::sha::{Impl, Sha256};
+
+    let mut bytes = [0u8; 64];
```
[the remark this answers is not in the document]
3esmit: @Copilot 
False positive. `Sha256` is a trait, and `Impl::hash_bytes` is one of its
methods — Rust requires the trait to be in scope to call it via the impl
type. Dropping `Sha256` from the `use` fails the build with:

```
error[E0599]: no function or associated item named `hash_bytes` found
              for struct `risc0_zkvm::sha::Impl` in the current scope
  --> stablecoin/core/src/lib.rs:95:31
   |
95 |     out.copy_from_slice(Impl::hash_bytes(&bytes).as_bytes());
   |                               ^^^^^^^^^^ function or associated item
   |                                          not found in `risc0_zkvm::sha::Impl`
   |
   = help: items from traits can only be used if the trait is in scope
help: trait `Sha256` which provides `hash_bytes` is implemented but not
      in scope; perhaps you want to import it
   |
 3 + use risc0_zkvm::sha::Sha256;
   |
```

The identical `use risc0_zkvm::sha::{Impl, Sha256};` pattern is used in
`amm/core/src/lib.rs` (four call sites) and `ata/core/src/lib.rs` for the
same reason. No change here.

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+    program::{PdaSeed, ProgramId},
+};
 use serde::{Deserialize, Serialize};
 
-#[derive(Serialize, Deserialize)]
+// Domain-separation tags for the PDA seeds derived by the Stablecoin Program.
+// These bytes are part of the on-chain derivation scheme and must stay unchanged for
+// account-address compatibility.
+const POSITION_PDA_DOMAIN: [u8; 32] = *b"stablecoin::position::seed::v1\0\0";
+const POSITION_VAULT_PDA_DOMAIN: [u8; 32] = *b"stablecoin::position::vault::v1\0";
```
0x-r4bbit: I like this, but I think we should stay consistent across programs on what the bytes look like (the AMM program uses non-descriptive bytes).
Maybe worth an LIP?

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+
+/// Stablecoin Program Instruction.
+#[derive(Debug, Serialize, Deserialize)]
 pub enum Instruction {
+    /// Heartbeat / sanity-check entry point that returns the input account unchanged.
     Noop,
```
0x-r4bbit: I think we can remove `Noop` as part of this entirely.

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+
+    /// Open a new collateral-only [`Position`] for the calling owner.
+    ///
+    /// Required accounts (5):
+    /// - Owner account (authorized)
+    /// - Position account (uninitialized, address must match
+    ///   `compute_position_pda(stablecoin_program_id, owner)`)
+    /// - Position vault token holding account (uninitialized, address must match
+    ///   `compute_position_vault_pda(stablecoin_program_id, position_id)`)
```
0x-r4bbit: One thing to check: are positions single-collateral? Is one holding account enough here? Should we otherwise use the token definition id as part of the PDA to make them unique?

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+    /// - Owner's source token holding for the collateral (authorized, initialized)
+    /// - Token definition account for the collateral (matches the user holding's `definition_id`;
+    ///   its `program_owner` determines the Token Program used by the chained `InitializeAccount`
+    ///   / `Transfer` calls)
+    OpenPosition {
+        /// `ProgramId` under which the [`Position`] and vault PDAs are derived.
+        stablecoin_program_id: ProgramId,
```
0x-r4bbit: I think we can use `self_program_id` via `ProgramContext` here (landed in `main`)

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+        collateral_amount: u128,
+    },
+}
+
+/// Persistent state held by a Stablecoin [`Position`] account.
+///
+/// `debt_amount` is included for forward compatibility with `generate_debt`; until that
+/// instruction lands `open_position` always initializes it to `0`.
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+pub struct Position {
```
0x-r4bbit, line 41: We should add the `[#account_type]` directive here

stablecoin/core/src/lib.rs
```diff
@@ -1,6 +1,169 @@
...
+        #[allow(
+            clippy::expect_used,
+            reason = "BorshSerialize::serialize is infallible when writing to a Vec"
+        )]
+        BorshSerialize::serialize(position, &mut data)
+            .expect("Serialization to Vec should not fail");
+        #[allow(
+            clippy::expect_used,
+            reason = "Position encodes to a small, bounded byte length that fits Data"
+        )]
```
0x-r4bbit: These `allow` directives are not needed at the moment because `clippy::expect_used` is not enabled in our project. Let's remove them, as they just add noise at this point.

stablecoin/methods/guest/src/bin/stablecoin.rs
```diff
@@ -1,19 +1,76 @@
...
+//!
+//! Wires the host-side `stablecoin_program` instruction handlers to the LEZ framework via
+//! `#[lez_program]` so the entry points can be invoked from a deployed program.
+
 #![no_main]
+#![allow(
+    missing_docs,
+    reason = "lez_program / instruction proc macros emit module, function, and constant items \
+              that do not carry doc strings; user-written handlers document inline"
+)]
```
0x-r4bbit: No need for this `allow` directive.

stablecoin/methods/guest/src/bin/stablecoin.rs
```diff
@@ -1,19 +1,76 @@
+//! RISC Zero guest binary for the Stablecoin Program.
+//!
+//! Wires the host-side `stablecoin_program` instruction handlers to the LEZ framework via
+//! `#[lez_program]` so the entry points can be invoked from a deployed program.
```
0x-r4bbit: To stay consistent with other programs, let's not add this comment

stablecoin/methods/guest/src/bin/stablecoin.rs
```diff
@@ -1,19 +1,76 @@
...
+                  literals into execute_with_claims; this handler delegates to a host function \
+                  that returns Vec<AccountPostState>, so migration requires restructuring the \
+                  handler shape — workspace-wide follow-up across token/amm/ata"
+    )]
     pub fn noop(account: AccountWithMetadata) -> SpelResult {
         Ok(SpelOutput::states_only(stablecoin_program::noop::noop(
             account,
         )))
     }
+
```
0x-r4bbit: We should remove `noop` as part of this PR.

stablecoin/methods/guest/src/bin/stablecoin.rs
```diff
@@ -1,19 +1,76 @@
...
+    /// # Errors
+    /// Returns the host program's panic-converted error if any precondition fails (see
+    /// [`stablecoin_program::open_position::open_position`] for the full list).
+    #[instruction]
+    #[allow(
+        deprecated,
+        reason = "SpelOutput::with_chained_calls: same reason as noop above — migration to \
+                  SpelOutput::execute requires the macro's vec![...] literal shape, which \
+                  conflicts with delegating to host helpers"
+    )]
```
0x-r4bbit: The deprecation allowance can be removed, because we're sitting on a newer version of spel, which uses `::execute()`. Please update your PR accordingly

stablecoin/methods/guest/src/bin/stablecoin.rs
```diff
@@ -1,19 +1,76 @@
...
+        let (post_states, chained_calls) = stablecoin_program::open_position::open_position(
+            owner,
+            position,
+            vault,
+            user_holding,
+            token_definition,
+            stablecoin_program_id,
+            collateral_amount,
+        );
+        Ok(SpelOutput::with_chained_calls(post_states, chained_calls))
```
0x-r4bbit: This should be using `SpelOutput::execute`

stablecoin/src/open_position.rs
```diff
@@ -0,0 +1,132 @@
...
+/// - `position` or `vault` is already initialized.
+/// - `position.account_id` / `vault.account_id` do not match their PDA derivations.
+/// - `user_holding` cannot be decoded as a [`TokenHolding`].
+/// - `user_holding`'s definition does not match `token_definition`.
+/// - `token_definition.program_owner` does not match `user_holding.program_owner`.
+#[allow(
+    clippy::needless_pass_by_value,
+    reason = "instruction handler shape: spel #[instruction] macro deserializes owned \
+              AccountWithMetadata values into these parameters"
+)]
```
0x-r4bbit: Why is this needed?

stablecoin/src/open_position.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    let position_seed = verify_position_and_get_seed(&position, &owner, stablecoin_program_id);
+    let vault_seed =
+        verify_position_vault_and_get_seed(&vault, position.account_id, stablecoin_program_id);
+
+    #[allow(
+        clippy::expect_used,
+        reason = "open_position uses the same panic-on-bad-input pattern as the surrounding \
+                  assert!/assert_eq! invariant checks; runtime must supply a valid TokenHolding"
+    )]
```
0x-r4bbit: As mentioned above, we don't used clippy::expect_used

stablecoin/src/open_position.rs
```diff
@@ -0,0 +1,132 @@
...
+        collateral_vault_id: vault.account_id,
+        collateral_definition_id: token_definition.account_id,
+        collateral_amount,
+        debt_amount: 0,
+    });
+
+    let post_states = vec![
+        AccountPostState::new(owner.account),
+        AccountPostState::new_claimed(position_post, Claim::Pda(position_seed)),
+        AccountPostState::new(vault.account.clone()),
```
0x-r4bbit, line 85: I think we need to claim the position vault via PDA seed as well, no?

0x-r4bbit (changes requested):

stablecoin/src/open_position.rs
3esmit: Because locally I am running a more agressive lint/clippy rules thorugh command line, which triggers that clippy false positive, I should have removed that but I missed it was just on my run that was needed.

stablecoin/core/src/lib.rs
3esmit: Done. Removed `Noop` from the core instruction enum, guest wrapper, program module, unit test, and regenerated IDL.

stablecoin/core/src/lib.rs
3esmit: The position PDA now includes both the owner account id and the collateral token definition id, so positions are unique per `(owner, collateral definition)` instead of one global position per owner.

stablecoin/core/src/lib.rs
3esmit: The guest now takes `ProgramContext` and passes `ctx.self_program_id` into the host handler. `stablecoin_program_id` is no longer part of the public instruction args or IDL.

stablecoin/methods/guest/src/bin/stablecoin.rs
3esmit: The guest now uses `SpelOutput::execute(post_states, chained_calls)` and no longer needs the deprecation allowance.

stablecoin/src/open_position.rs
3esmit: Removed the local `clippy::needless_pass_by_value` allowance.

stablecoin/src/open_position.rs
3esmit: Removed the extra `clippy::expect_used` allowances from the core serialization path and `open_position`.

stablecoin/src/open_position.rs
3esmit: Hmm.... The vault claim is intentionally handled by the chained Token `InitializeAccount` call, not by the Stablecoin post-state. Stablecoin passes the vault PDA seed with `.with_pda_seeds(vec![vault_seed])`, which authorizes the Token program to claim and initialize the vault as a token holding. Claiming it in Stablecoin's own post-states would make the Stablecoin program the vault owner, which is not what we want for a Token holding.

3esmit: I added a source comment to make that boundary explicit.

stablecoin/methods/guest/src/bin/stablecoin.rs
3esmit: The guest now uses `SpelOutput::execute(post_states, chained_calls)` and no longer needs the deprecation allowance.

stablecoin/core/src/lib.rs
3esmit: Replaced the descriptive Stablecoin PDA domain strings with numeric `[0; 32]` and `[1; 32]` tags so this follows the AMM style for now. Agree a LIP would be the right place to standardize a cross-program PDA seed convention.

stablecoin/core/src/lib.rs
3esmit: Added `#[account_type]` to `Position` and regenerated `artifacts/stablecoin-idl.json`.

stablecoin/core/src/lib.rs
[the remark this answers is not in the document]
3esmit: The `Sha256` trait import is now written as `Sha256 as _`, which keeps the trait in scope for `Impl::hash_bytes` without looking like a normal unused import.

stablecoin/core/src/lib.rs
3esmit: Removed the extra `clippy::expect_used` allowances from the core serialization path and `open_position`.

stablecoin/methods/guest/src/bin/stablecoin.rs
3esmit: Removed the Stablecoin guest crate-level docs block and the crate-level `missing_docs` allowance.

stablecoin/methods/guest/src/bin/stablecoin.rs
3esmit: Sure!

stablecoin/src/open_position.rs
0x-r4bbit: Thank you for clarifying. I still get confused by this.

0x-r4bbit (approved):

PR #109, opened 2026-05-14
feat(stablecoin): add price feed interface

3esmit:
Closes #96

## Summary

Adds the documented price-feed interface as a shared oracle surface and introduces a Mock Oracle Program that can write canonical price accounts for tests and development.
The mock oracle has no upstream oracle integration: it only materializes the supplied mocked price into the same account shape that consumers read.

This does not implement a TWAP oracle, RedStone adaptor, or stablecoin rate-update instruction.

## What Changed

- Added `oracle_core` with the canonical `OraclePriceAccount` fields:
  `base_asset`, `quote_asset`, `price`, `timestamp`, `source_identifier`, and
  `confidence_interval`.
- Added the consumer-side `PriceFeed` interface, `AccountPriceFeed`,
  `PriceFeedConfig`, `PriceFeedReading`, `PriceFeedError`, and shared
  `validate_price_account` checks.
- Re-exported the oracle price-feed surface from `stablecoin_core` so existing
  stablecoin consumers can use the interface without duplicating oracle types.
- Added the `mock_oracle` program family with a `set_price` instruction that writes a supplied mock price into an authorized price account owned by the Mock Oracle Program.
- Added `artifacts/mock_oracle-idl.json` and regenerated `artifacts/stablecoin-idl.json`.
- Updated workspace manifests and guest locks for `oracle_core` and `mock_oracle`.
- Updated two ATA integration tests to match the current `ATA::Transfer` instruction shape so the full integration test gate compiles.

## Justification

Issue #96 requires a configurable feed interface that can read market prices from an external feed account, enforce a caller-selected freshness threshold, and return errors when the feed is unavailable, stale, or inconsistent with the expected asset pair.

Keeping the canonical account and consumer checks in `oracle_core` lets real oracle producers and mock feeds write the same account shape while each consuming program owns its validation policy. The Mock Oracle Program intentionally avoids any real oracle verification and only provides a controlled way to set prices through the public program interface.

## Risks

- Price scaling is intentionally not fixed here. Future controller code must document the chosen precision before using `PriceFeedReading::price` in financial math.
- Timestamp units are consumer-owned: `timestamp`, `current_timestamp`, and `max_age` must use the same unit.
- The mock oracle accepts caller-supplied prices by design and should not be treated as a production oracle source.

## Validation

```bash
taplo fmt --check .
git diff --check
cargo +1.94.0 build -p idl-gen --release
# full IDL drift loop from AGENTS.md
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3045,6 +3045,7 @@ dependencies = [
  "risc0-zkvm",
  "serde",
  "spel-framework-macros",
+ "twap_oracle_core",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


0x-r4bbit (approved):

0x-r4bbit:
Done!

PR #110, opened 2026-05-18
chore: update spel to v0.3.0

0x-r4bbit:


3esmit (approved):
This PR just changes Cargo.toml and Cargo.lock files. I haven't tested integration wise on this change, but it should not break anything as no breaking  changes implemented on this Spel update.

PR #112, opened 2026-05-19
feat(stablecoin): implement `withdraw_collateral` (#92)

gravityblast:
closes #92
closes #93
closes #94

Note: fee accrual and post-withdrawal collateralization checks are defered to be implemented together with generate_debt.

### DIFF
--- a/stablecoin/src/tests.rs
+++ b/stablecoin/src/tests.rs
@@ -30,6 +30,26 @@ fn user_holding_id() -> AccountId {
     AccountId::new([0x30u8; 32])
 }
 
+fn token_holding_account(
+    account_id: AccountId,
+    definition_id: AccountId,
+    balance: u128,
+) -> AccountWithMetadata {
+    AccountWithMetadata {
+        account: Account {
+            program_owner: TOKEN_PROGRAM_ID,
+            balance: 0,
+            data: Data::from(&TokenHolding::Fungible {
+                definition_id,
+                balance,
+            }),
+            nonce: Nonce(0),
+        },
+        is_authorized: false,
+        account_id,
+    }
+}
+
 fn position_id() -> AccountId {
     compute_position_pda(
         STABLECOIN_PROGRAM_ID,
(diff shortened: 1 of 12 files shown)


stablecoin/src/tests.rs
```diff
@@ -392,3 +446,268 @@ fn position_pda_and_vault_pda_do_not_collide() {
...
 }
+
+#[test]
+fn withdraw_collateral_updates_position_and_emits_transfer() {
+    let initial_collateral: u128 = 500;
+    let amount: u128 = 200;
+    let (post_states, chained_calls) = crate::withdraw_collateral::withdraw_collateral(
+        owner_account(),
+        init_position_account(initial_collateral, 0),
+        init_vault_account(),
```
3esmit, line 479: Vault account is initialized with balance 0

stablecoin/src/tests.rs
```diff
@@ -392,3 +446,268 @@ fn position_pda_and_vault_pda_do_not_collide() {
     let vault = compute_position_vault_pda(STABLECOIN_PROGRAM_ID, position);
     assert_ne!(position, vault);
 }
+
+#[test]
+fn withdraw_collateral_updates_position_and_emits_transfer() {
+    let initial_collateral: u128 = 500;
+    let amount: u128 = 200;
```
3esmit, line 475: Withdraw requests for 200

stablecoin/src/tests.rs
```diff
@@ -392,3 +446,268 @@ fn position_pda_and_vault_pda_do_not_collide() {
...
+    vault_authorized.is_authorized = true;
+    let expected_transfer = ChainedCall::new(
+        TOKEN_PROGRAM_ID,
+        vec![vault_authorized, destination_holding_account()],
+        &token_core::Instruction::Transfer {
+            amount_to_transfer: amount,
+        },
+    )
+    .with_pda_seeds(vec![compute_position_vault_pda_seed(position_id())]);
+    assert_eq!(chained_calls[0], expected_transfer);
```
3esmit, line 521: Only checks if `Transfer` is emmited, but the downstream token program would reject the transfer because it does not have balance.

3esmit (changes requested):
Add a test that gives the vault a balance matching `Position.collateral_amount` and either executes the emitted transfer through `token_program::transfer::transfer` or, better, adds an integration test that opens a stablecoin position and then withdraws from it through LEZ chained-call execution

PR #130, opened 2026-05-28
feat(twap-oracle): implement CreateOraclePriceAccount instruction

0x-r4bbit:
Adds the CreateOraclePriceAccount instruction to the TWAP oracle program. The instruction initialises a canonical OraclePriceAccount PDA for a given price source and time window. The account starts with price = 0 and timestamp = 0 — a deliberately invalid sentinel state that signals "not yet published". Consumers are expected to reject any account whose timestamp is zero or stale, so the transient invalid state requires no special on-chain enforcement.

- PDA mirrors PriceObservations: derived from (oracle_program_id, price_source_id, window_duration) with a distinct seed constant, so each (source, window) pair maps to a distinct oracle price account that cannot collide with its corresponding observations account.

- source_id is not a parameter: it is always set to price_source.account_id. Accepting it as a free parameter would allow callers to register a price account that claims to represent a source it does not control. Deriving it from the authorized price source account closes that vector entirely.

- Authorization follows the same model as CreatePriceObservations: is_authorized = true on the price source proves the caller controls it; the PDA check ensures the supplied oracle price account address is the one derived from that specific source and window.

- price = 0 / timestamp = 0 is the correct initial state: coupling account creation to first publication would require the observation account to already hold a full window of ticks, blocking registration for up to the window duration. Consumers must validate oracle prices regardless, so the zero sentinel falls naturally out of the staleness check they already own.

Closes #129

**This needs #128 to land first**

### DIFF
--- a/programs/twap_oracle/src/create_oracle_price_account.rs
+++ b/programs/twap_oracle/src/create_oracle_price_account.rs
@@ -1,0 +1,524 @@
+use clock_core::{ClockAccountData, CLOCK_01_PROGRAM_ACCOUNT_ID};
+use nssa_core::{
+    account::{Account, AccountId, AccountWithMetadata, Data},
+    program::{AccountPostState, Claim, ProgramId},
+};
+use twap_oracle_core::{
+    compute_oracle_price_account_pda, compute_oracle_price_account_pda_seed, OraclePriceAccount,
+};
+
+/// Creates and initialises an [`OraclePriceAccount`] for a price source account and time window.
+///
+/// The account is initialised with a non-zero `initial_price` and the timestamp read from the
+/// canonical 1-block LEZ clock. Both must be non-zero: a zero price or zero timestamp is the
+/// sentinel consumers treat as "no valid price", so an account must never be created in that
+/// state. `confidence_interval` starts at zero (the source may not provide one).
+///
+/// `initial_price` is a `Q64.64` fixed-point value: the real price is `initial_price / 2^64`, so
+/// `1.0` is `1 << 64`. The non-zero check rejects the sentinel but cannot validate scale — a
+/// caller is responsible for supplying a correctly-scaled fixed-point price, not a plain integer.
+///
+/// The timestamp is taken from `clock`, which must be [`CLOCK_01_PROGRAM_ACCOUNT_ID`]; it is never
+/// caller-supplied, so it cannot be forged.
+///
+/// Authorization is implicit in the PDA relationship: the oracle price account is derived from
+/// `price_source.account_id` and `window_duration`, so whoever controls the price source
+/// controls this account.
+///
+/// # Panics
+/// Panics if:
+/// - `oracle_price_account.account_id` does not match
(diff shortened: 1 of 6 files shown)


programs/twap_oracle/src/create_oracle_price_account.rs
```diff
@@ -0,0 +1,369 @@
...
+        compute_oracle_price_account_pda(oracle_program_id, price_source_id, window_duration),
+        "CreateOraclePriceAccount: oracle price account ID does not match expected PDA"
+    );
+    assert_eq!(
+        oracle_price_account.account,
+        Account::default(),
+        "CreateOraclePriceAccount: oracle price account must be uninitialized"
+    );
+    assert!(
+        price_source.is_authorized,
```
3esmit, line 64: Is this correct? Any account that can be authorized by the caller can be used as `price_source`.

programs/twap_oracle/src/create_oracle_price_account.rs
```diff
@@ -0,0 +1,369 @@
...
+        "CreateOraclePriceAccount: price source account must be authorized (caller must control it via a PDA)"
+    );
+
+    let account = OraclePriceAccount {
```
3esmit, line 83: This should be an invalid OraclePriceAccount, because RFP sayst that interface must reject zero, negative, or invalid prices. Could it be initialized with the correct first time values?

3esmit (dismissed):

programs/twap_oracle/src/create_oracle_price_account.rs
0x-r4bbit: Yes, that's the intention. The `price_source` is an account that authorizes the creation of a price account derived from that price source. Example:

- You can create an account (price source) and create a price account for it (e.g. 24 hrs)
- The amm's pool is a price source, and it can also create a price account for it

You cannot create a price account for a price source that has not authorized that creation.

programs/twap_oracle/src/create_oracle_price_account.rs
0x-r4bbit: This is something were we might want to fine-tune the RFP.
Two options here:

1. We require values to be non-zero, but that also means before creating a price account, you need to have a price you can actually store. In case of a price that comes from an observation account (like via the amm), this means the observation account has to exist first **and** it needs to have at least two values recorded, which, depending on the time window could drag the creation of the price accounts.

2. We allow for zero values, and assume that consumers do validation on their side (which they have to do anyways). Meaning, if the price is `0` or the `last_updated` timestamp is older than the time window for the price account, the price can be considered invalid and should be rejected. In either case, anyone can update the price, which should make it `> 0` and removes the stale `last_udpated` value.

I went with option 2 now.

0x-r4bbit: As per our discussion, I went with 1) now

gravityblast (approved):

PR #133, opened 2026-06-01
Add DepositCollateral

3esmit:
## Summary

- Add `DepositCollateral` to the stablecoin instruction surface and guest entrypoint.
- Move additional collateral from a user holding into the existing position vault through a chained `token_core::Instruction::Transfer`.
- Anchor the chained transfer target to the collateral token definition's Token Program, and reject mismatched vault/source owners.
- Require both source and vault collateral holdings to be fungible before increasing `Position.collateral_amount`.
- Skip zero-amount downstream transfers while keeping account validation and post-state output.
- Increase `Position.collateral_amount` without a collateralization check, since debt is unchanged.
- Mark the mutable `deposit_collateral` IDL accounts as writable and update unit/integration coverage.

Closes #90

### DIFF
--- a/artifacts/stablecoin-idl.json
+++ b/artifacts/stablecoin-idl.json
@@ -49,23 +49,29 @@
         {
           "name": "owner",
           "writable": false,
-          "signer": false,
+          "signer": true,
           "init": false
         },
         {
           "name": "position",
-          "writable": false,
+          "writable": true,
           "signer": false,
           "init": false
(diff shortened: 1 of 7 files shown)


gravityblast:
@0x-r4bbit @3esmit I think this PR is still based on an old plan, can we close it ?

3esmit:
Yes

PR #141, opened 2026-06-08
docs(stablecoin): add stablecoin design docs

gravityblast:


stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**Time passes; debt grows; users adjust.** As keepers advance the fee multiplier, every position's debt grows even though no per-position write happens (the RAI accumulator approach, §6.1). Users can:
+
+- `deposit_collateral` (§10.5) to add more collateral and improve the ratio.
+- `withdraw_collateral` (§10.6) to take some back — only allowed if the position still meets the safety ratio afterwards.
+- `repay_debt` (§10.8) to burn stablecoin and lower the debt. To pay off accrued interest, users may need to buy extra stablecoin on the market.
+
+**Closing out.** When a position's debt and collateral are both zero, `close_position` (§10.9) clears the `Position` account so the user can later open a new position with the same nonce. (The vault account stays — see §14.)
```
3esmit: Please, clarify about nonce reuse. Later `11. edge-case` sections says same-nonce reopen fails because the vault PDA lingers. Close position would always free the nonce, or there are some specific casess where it dosent?

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+error: i256 = (current_redemption_price as i256) - (oracle.price as i256)
+
+// 3. Update integral state (clamped for anti-windup).
+integral_delta = (params.controller_integral_gain as i256) * error * (dt as i256) / FIXED_POINT_ONE
+new_integral = state.controller_integral_term + integral_delta
+new_integral = clamp(new_integral, -INTEGRAL_CLAMP, INTEGRAL_CLAMP)
+
+// 4. Compute rate adjustment.
+proportional_term = (params.controller_proportional_gain as i256) * error / FIXED_POINT_ONE
+rate_adjustment   = -(proportional_term + new_integral)
```
3esmit: This whole math should be reviewed, its very confusing. 

Line 464 `error = current_redemption_price - oracle.price` //    error > 0 when redemption_price > market_price (protocol's target above market valuation).

467: `integral_delta` carry this signal as the difference of past integral term;
468: `new_integral` sums the `integral delta`

472: `proportional_term` carry error signal (so its negative when below market valuation, positive when above)
473 `rate_adjustment` sum prortional_term and the new  integral, than inverts signal` 


Inverting the signal at rate_adjustment does not make sense for me, because: `proportional_term` was already on the correct signall; and new_integral is the whole value. 

Please verify this logic.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+    PP -. id ref .-> Oracle
+    Pos -. id ref .-> Vault
+    Vault -. holds .-> CollDef
+```
+
+### How it all works
+
+A walkthrough of who calls what and what changes when. Detailed mechanics are in §10; this is the overall flow.
+
+**Day 0 — setup.** The deployer publishes the program binary, then calls `initialize_program` (§10.1). That one call creates all five program-owned accounts: the stablecoin `TokenDefinition` (via a chained `Token::NewFungibleDefinition`), its required paired master holding (empty), the `ProtocolParameters` account, and the two state-tracking accounts (`StabilityFeeAccumulator` and `RedemptionPriceState`). The same call sets which collateral and oracle the protocol uses, picks the admin and freeze-authority accounts, and stores the initial fee rate, controller gains, safety ratio, and timing parameters. Once it returns, the protocol exists and users can start interacting.
```
0x-r4bbit, line 124: > The deployer publishes the program binary, then calls `initialize_program`

Leaving this here to make sure we add this to the doc in case it's missing:
- `initialize_program` can be front-run and we don't have `initializeAndCall()` equivalents (from EVM)
- Hence, we'll have to hardcode an authority account that is permitted to initialize the program
- That account cannot be changed (as we don't have an upgrade solution yet) (or, maybe one other solution to this would be to also have a primary and secondary authority account. The secondary is the hardcoded one and will be used when the primary is not set. The primary one will be set by the secondary one, allowing for changing the authority later on.

@3esmit @gravityblast what do you guys think?

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**In plain English:** this is just `rate ^ seconds_elapsed` — what a per-second multiplier becomes after that many seconds. The naive way is `seconds_elapsed` separate multiplications. Exponentiation-by-squaring does it in `log₂(seconds_elapsed)` multiplications instead — for a year's worth of seconds (~31.5M), that's ~25 muls instead of 31.5M.
+
+### 5.3 Current value projections (read on the hot path)
+
+```rust
+// Used in every op that needs the up-to-date accumulator / redemption price.
+fn current_accumulated_rate(state: StabilityFeeAccumulator, params: ProtocolParameters, now: u64) -> u128 {
+    let dt = now.saturating_sub(state.last_accrued_at);
+    let factor = compound_rate(params.stability_fee_per_second, dt);
```
3esmit: If  `stability_fee_per_second = FIXED_POINT_ONE * 2`, and a keeper asks `compound_rate` to compute `2^86400`, which cannot fit in `u128`.

Should `dt` need a max elapsed window? Or some other design?

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**Running the protocol — keepers advance the globals.** Two instructions any account can call keep the protocol's globals current:
+
+- `accrue_stability_fee` (§10.2) advances the global fee multiplier. Nothing moves; the multiplier grows so every open position accrues interest at once.
+- `update_redemption_rate` (§10.3) reads the market-price oracle, runs the PI controller (§6.4), and updates the redemption price and its drift rate. The redemption price then keeps drifting toward the market price until the next call.
+
+These two are usually run by keeper bots that batch them into one transaction. They cost gas but earn no reward — anyone with money in the protocol wants the globals fresh before they act.
```
0x-r4bbit: Maybe we can have a convenience `update_rates_and_fees` function that does both of these.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**Outputs:**
+
+- `position.normalized_debt_amount` ← `old − ⌊amount × FIXED_POINT_ONE / current_accumulator⌋` (round DOWN, § 6.3); `checked_sub` panics on overrepay.
+- `stablecoin_definition.total_supply` ← `old − amount`.
+- `user_stablecoin_holding.balance` ← `old − amount`.
+
+**Chained calls:** `Token::Burn { amount_to_burn: amount }` · accounts: `[stablecoin_definition, user_stablecoin_holding (user-authorized)]` · no PDA seeds.
+
+**Panics if:** `owner.is_authorized = false`; `user_stablecoin_holding.is_authorized = false`; position uninit / wrong owner / PDA mismatch; `stablecoin_definition.account_id ≠ protocol_parameters.stablecoin_definition_id`; user holding's `definition_id` mismatch or different Token Program; overrepay (`⌈amount × FIXED_POINT_ONE / current_accumulator⌉ > position.normalized_debt_amount`). Allowed when frozen.
```
3esmit: overrepay (`⌈amount × FIXED_POINT_ONE / current_accumulator⌉ > position.normalized_debt_amount`

On line 896 it uses `⌊`, and in here is uses `⌈`.  Which is the correct formula?

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+- `update_redemption_rate` (§10.3) reads the market-price oracle, runs the PI controller (§6.4), and updates the redemption price and its drift rate. The redemption price then keeps drifting toward the market price until the next call.
+
+These two are usually run by keeper bots that batch them into one transaction. They cost gas but earn no reward — anyone with money in the protocol wants the globals fresh before they act.
+
+**A user opens a position and borrows.** A user picks a `position_nonce` (any integer — 0, 7, whatever) and calls `open_position` (§10.4) with some collateral. Two accounts get created: their `Position` (at `hash(owner, nonce)`, owned by the stablecoin program, starting with zero debt) and a `PositionVault` (at `hash(position)`, owned by the Token Program) that holds the collateral tokens. They can then call `generate_debt` (§10.7) to mint stablecoin into their own holding. The mint increases the position's normalized debt; the collateralization check prevents over-borrowing; the call fails if the oracle is stale or the protocol is frozen.
```
0x-r4bbit, line 133: >  A user picks a `position_nonce` (any integer — 0, 7, whatever)

This is likely to allow for multiple positions per account (and maybe even becomes part of the PDA).
This paragraph states that it's user controlled and I think that's generally fine as long as there's constraints that an existing nonce hasn't already been used.

However, something has to keep track of the user's nonces somewhere. Maybe we can think of approaches/recommendations on where/how these are tracked.

stablecoin/docs/README.md
0x-r4bbit: Note:  Define what the constraints are to distinguish between a used and unused vault with the same nonce

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+## 9. Instruction set
+
+18 instructions in 5 groups. Full per-instruction details in § 10.
+
+### 9.1 Bootstrap
+
+| # | Instruction | Caller | One-line |
+|---|---|---|---|
+| 1 | `initialize_program` | admin (one-shot) | Create all four global PDAs (three native + stablecoin definition via chained `Token::NewFungibleDefinition`), bind collateral + oracle + initial params, set initial redemption price. |
```
3esmit: This says four target PDAs, but `initialize_program` describes five claimed accounts (line 121, 146)

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**`RedemptionPriceState`** — PDA seed `"REDEMPTION_PRICE_STATE"`, owned by the stablecoin program.
+
+*Holds:* the redemption price anchor (at the last update), the per-second drift rate, the PI controller's persisted integral term, and the wall-clock timestamp of the last update.
+
+*Why:* the redemption price is the protocol's target value (in collateral-per-stablecoin); the controller continuously drifts it based on the deviation between market price and target. Storing as `(anchor, rate, last_updated_at)` lets the current value be projected on the fly without writing every block — only `update_redemption_rate` re-anchors it.
+
+**`StablecoinDefinition`** — PDA seed `"STABLECOIN_DEFINITION"`, owned by the **Token Program** (PDA-derived under the stablecoin program).
+
+*Holds:* the stablecoin's `TokenDefinition::Fungible` — name, current `total_supply`, no metadata.
```
0x-r4bbit: I think in terms of metadata, we should at least have a name here, no?

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**`StablecoinMasterHolding`** — PDA seed `"STABLECOIN_MASTER_HOLDING"`, owned by the **Token Program** (PDA-derived under the stablecoin program).
+
+*Holds:* a `TokenHolding::Fungible` for the stablecoin with `balance = 0` permanently.
+
+*Why this exists at all:* a Token-Program-API artifact. `Token::NewFungibleDefinition` always creates a definition AND a paired holding in the same call, and mints `total_supply` units into that holding. There's no variant that creates a definition alone. We pass `total_supply = 0` — our stablecoin starts with **zero** supply because every coin in existence must come from a user calling `generate_debt` against real collateral (a non-zero initial supply would be free, unbacked money). The master holding is therefore born empty and never touched again. Modeling it as a deterministic stablecoin-program PDA contains the artifact at a known, addressable location instead of leaking it into someone's user account.
```
0x-r4bbit, line 179: **Actually**: I wonder if this is something that should be changed. I see that it's convenient to create a holding account when also creating a definition, but if there's cases where one doesn't necessarily want to create a holding right away, I think that should be possible too.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+A future Token Program extension (`Token::NewFungibleDefinitionWithoutHolding`, tracked in §14 follow-ups) would let `initialize_program` drop this account entirely.
```
0x-r4bbit, line 181: Ha, ignore my comment aboth. Although, I don't think this should be a new instruction.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**Why split the globals?** Each one has exactly one writer family:
+
+- `ProtocolParameters` ← `initialize_program`, admin `set_*`, `freeze` / `unfreeze`.
+- `StabilityFeeAccumulator` ← `initialize_program`, `accrue_stability_fee`, `set_stability_fee_per_second` (auto-accrues inline).
+- `RedemptionPriceState` ← `initialize_program`, `update_redemption_rate`.
+- `StablecoinDefinition` + `StablecoinMasterHolding` ← `initialize_program` via chained `Token::NewFungibleDefinition`; afterwards the definition's `total_supply` is incremented / decremented by `Token::Mint` / `Token::Burn` from `generate_debt` / `repay_debt`. The master holding is never touched again.
+
+That strict writer separation means a keeper calling `accrue_stability_fee`, another keeper calling `update_redemption_rate`, and a user calling `withdraw_collateral` in the same block don't contend on any account.
```
0x-r4bbit: This is a good section.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+```
+
+### 4.4 `Position`
+
+```rust
+#[account_type]
+pub struct Position {
+    /// Owner of the position. Required to be `is_authorized` for every position op.
+    /// Also stored for client discovery (PDA seed isn't reversible).
+    pub owner_account_id: AccountId,
```
0x-r4bbit, line 312: So far, I've omitted such redundant information in the oracle account structures (since it lives in the PDAs, so you need to know that ID to find the account, which means you already have it).

However, it doesn't address cases where one finds/looks at an account and wants to figure out who the owner is. In this case attaching the ID is necessary.

What do you guys generally think about this @3esmit @gravityblast

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+### 3.2 Per-position accounts (one set per open position)
+
+Created on demand by `open_position` and cleared by `close_position`. Both accounts' addresses are deterministic from the owner and a nonce, so clients can discover positions without an off-chain index.
+
+**`Position`** — PDA seed `hash(owner_account_id, position_nonce)`, owned by the stablecoin program.
+
+*Holds:* owner id, position nonce, the vault PDA address, current collateral amount, normalized debt amount, and the opened-at timestamp.
+
+*Why:* the canonical state for one CDP. Owner authorization is checked against `owner_account_id` on every op. Storing the vault id explicitly is redundant (it's derivable) but saves the derivation cost on every read. The whole struct is owned by the stablecoin program — that's how we restrict who can write to it (only this program's instructions).
```
0x-r4bbit, line 200: > Storing the vault id explicitly is redundant (it's derivable) but saves the derivation cost on every read.

Which derivation cost? We can't derive anything from a PDA

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+**`PositionVault`** — PDA seed `hash(position_id)`, owned by the **Token Program** (PDA-derived under the stablecoin program).
+
+*Holds:* a `TokenHolding::Fungible` for the collateral token, with `balance` mirroring `Position.collateral_amount` after every op.
```
0x-r4bbit, line 204: Maybe can list invariants somewhere, where this is one of the items.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+### 4.1 `ProtocolParameters`
+
+```rust
+#[account_type]
+pub struct ProtocolParameters {
+    /// Authority required for every parameter-update and admin-rotation instruction.
+    pub admin_account_id: AccountId,
+    /// Authority required for `freeze` / `unfreeze`.
+    pub freeze_authority_account_id: AccountId,
```
0x-r4bbit, line 242: Maybe another sidequest would be to look into a `ROLES` implementation for LEZ

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+    /// Updatable by admin via `set_stability_fee_per_second` (auto-accrues first).
+    pub stability_fee_per_second: u128,
+    /// PI controller Kp. Signed.
+    pub controller_proportional_gain: i128,
+    /// PI controller Ki. Signed.
+    pub controller_integral_gain: i128,
+    /// Minimum collateralization ratio in fixed-point. e.g. `1.5 * FIXED_POINT_ONE` = 150%.
+    pub minimum_collateralization_ratio: u128,
+    /// Min seconds between successful `accrue_stability_fee` calls (spam guard).
+    pub minimum_seconds_between_fee_accruals: u64,
```
0x-r4bbit: Is calling `accrue_stability_fee` a lot a problem? Why would we need to guard against this (I'm probably missing something..)

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+    /// Min seconds between successful `update_redemption_rate` calls (RFP F2).
+    pub minimum_seconds_between_rate_updates: u64,
+    /// Reject oracle observations older than this (RFP R3 staleness gate).
+    pub maximum_oracle_price_age_seconds: u64,
```
0x-r4bbit: LEZ's block timestamps are in milliseconds. I've designed the TWAP oracle properties around this as well. So we should probably adjust here.

In fact, this applies to the `minimum_*` properties too

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+
+```rust
+#[account_type]
+pub struct StabilityFeeAccumulator {
+    /// Accumulator at `last_accrued_at`. Initialized to `FIXED_POINT_ONE`.
+    /// Updated on accrual via `anchor * compound_rate(stability_fee_per_second, Δt)`.
+    pub accumulated_rate_at_last_accrual: u128,
+    /// Unix seconds of the last accrue. Current accumulator on the read side =
+    /// `anchor * compound_rate(rate, now - last_accrued_at) / FIXED_POINT_ONE`.
+    pub last_accrued_at: u64,
```
0x-r4bbit, line 282: This should be ms too.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+### 4.3 `RedemptionPriceState`
+
+```rust
+#[account_type]
+pub struct RedemptionPriceState {
+    /// Redemption price at `last_updated_at`, in collateral per stablecoin, fixed-point.
+    pub redemption_price_at_last_update: u128,
+    /// Per-second drift multiplier, stored as `(1 + r) * FIXED_POINT_ONE`.
+    /// Below `FIXED_POINT_ONE` = decay. Output of the PI controller.
+    pub redemption_rate_per_second: u128,
```
0x-r4bbit: I wonder if we should adjust these or multiply/divide by 1000 as necessary

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+    /// Persisted integral state of the PI controller. Clamped on every update for
+    /// anti-windup (§ 6.4).
+    pub controller_integral_term: i128,
+    /// Unix seconds of the last update. Read-side current price =
+    /// `anchor * compound_rate(rate, now - last_updated_at) / FIXED_POINT_ONE`.
+    pub last_updated_at: u64,
```
0x-r4bbit, line 301: Same here, should be ms

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+### 4.4 `Position`
+
+```rust
+#[account_type]
+pub struct Position {
+    /// Owner of the position. Required to be `is_authorized` for every position op.
+    /// Also stored for client discovery (PDA seed isn't reversible).
+    pub owner_account_id: AccountId,
+    /// User-chosen on `open_position`. Together with owner forms the PDA seed.
+    pub position_nonce: u64,
```
0x-r4bbit, line 314: This is another such property that, in theory, could be ommitted.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+    /// Collateral vault PDA. Stored explicitly for op-time efficiency.
+    pub vault_account_id: AccountId,
+    /// Collateral atomic units. INVARIANT: equals `vault_holding.balance` after every op.
+    pub collateral_amount: u128,
+    /// Stablecoin atomic units divided by the accumulator at mint time.
+    /// **Nominal debt at time T** = `normalized_debt_amount * accumulated_rate(T) / FIXED_POINT_ONE`.
+    /// Storing normalized lets one global accumulator update apply interest to every position.
+    pub normalized_debt_amount: u128,
+    /// Unix seconds when the position was first opened. UX/analytics; not used in protocol logic.
+    pub opened_at: u64,
```
0x-r4bbit, line 324: ms

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+- Rounding direction is chosen per use site to favour the protocol (§ 6.5).
+
+### 5.2 `compound_rate`
+
+```rust
+/// Returns `per_second_rate ^ seconds_elapsed` in fixed point (where `1.0 == FIXED_POINT_ONE`).
+///
+/// O(log seconds_elapsed) — exponentiation by squaring. Uses u256 intermediates.
+/// Same algorithm as MakerDAO / RAI's `rpow`.
+pub fn compound_rate(per_second_rate: u128, seconds_elapsed: u64) -> u128;
```
0x-r4bbit: Should be using ms too?

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+// 6. Persist.
+state.redemption_price_at_last_update = current_redemption_price
+state.redemption_rate_per_second      = (FIXED_POINT_ONE as i256 + rate_adjustment) as u128
+state.controller_integral_term        = new_integral
+state.last_updated_at                 = now
+```
+
+The signs work out so that **positive gains drive the system toward stability**. Operators tune gain magnitude; sign is conventional and embedded in the controller, not the gain.
+
+`INTEGRAL_CLAMP` and `RATE_DELTA_CLAMP` are constants in v1 (§ 8). Promoting them to `ProtocolParameters` admin-tunable fields is a future revision (no on-chain migration needed — additive).
```
0x-r4bbit, line 499: How do we decide at what values we're clamping?

0x-r4bbit: This has been answered later in the document.

stablecoin/docs/README.md
```diff
@@ -0,0 +1,1269 @@
...
+- **RFP-001 (Admin Authority).** When RFP-001 ships, `admin_account_id` either points at the admin program's account (and any `set_*` instruction's `admin` input is that account being authorized by that program) or this RFP grows a wrapper. Either way it's a local change: `set_*` instructions become "admin authority program authorizes this account" rather than "this account is `is_authorized`". No data-model migration needed.
+- **RFP-002 (Freeze Authority).** Same pattern for `freeze_authority_account_id`.
+- **Mini-app / CLI / SDK.** Read the on-chain accounts directly to display: position-level collateralization, redemption price drift, projected outcomes (these are SDK functions over the same math in § 6). The deshield-interact-reshield privacy pattern is the SDK's responsibility; the program is unaware of whether callers are ephemeral or persistent.
+
+## 14. Open follow-ups (not blocking this design)
+
+- **`Token::CloseHolding`.** Upstream extension to the token program. Lets `close_position` actually clear the vault account so position-nonce reuse becomes possible. Track as a separate issue against the Token Program.
```
0x-r4bbit: What's the idea of this exactly?

stablecoin/docs/README.md
gravityblast: @0x-r4bbit I think a constructor or initialize function is needed in LEZ like in any other smart contract chain. it's even easier then `initializeAndCall` cause we are still not talking about upgradeability but just initialization on deployment. For this reason I would avoid hardcoding things, I would just expect to be able to call initialize in testnet without front running problems and have initialization built-in in production

0x-r4bbit: @schouhy Do you think we can have constructor-like capabilities for prod?

stablecoin/docs/README.md
gravityblast: the position and vault PDAs derivations have the nonce inside so if you use the same nonce those accounts will be already initialized, does it make sense @0x-r4bbit ?

0x-r4bbit: It does, I'm just thinking generally, for apps and other tools that might want to create and use these. They probably don't want to do trial and error until they find an available account.

0x-r4bbit (approved):

PR #142, opened 2026-06-08
Implement feedback controller for redemption rate

3esmit:
## Summary

- Add the stablecoin `RedemptionController` account, PDA helpers, and fixed-point gain scale.
- Add initialize/update instructions and guest handlers for permissionless redemption-rate updates.
- Compute signed P/PI redemption rates from oracle market price vs. redemption price using negative feedback, clamp integral state and rate, and pause updates when the configured feed is stale or unavailable.
- Bind the controller feed to stablecoin/collateral oracle assets without pinning the oracle account's program owner.
- Constrain controller instruction outputs to the supplied `current_timestamp` so oracle freshness is tied to execution time.

Closes #95.


gravityblast:
@0x-r4bbit @3esmit I think this PR is still based on an old plan, can we close it ?

3esmit:
Yes

PR #148, opened 2026-06-18
feat(amm): create TWAP price observations on behalf of the pool

0x-r4bbit:
Add a `CreatePriceObservations` instruction that registers a TWAP price-observations account for a pool over a time window, via a chained call to the configured TWAP oracle program. The pool acts as the price source: the AMM authorizes it with its pool PDA seed so the oracle ties the feed to that pool.

The feed's initial tick is read from the pool's authoritative `CurrentTickAccount` (validated against its pool-derived PDA) rather than being supplied by the caller, so the feed cannot be seeded at a forged price — mirroring what `RecordTick` does. The clock is verified to be the canonical 1-block LEZ clock, and creation is rejected if the observations account already exists.

To support the chained call, `AmmConfig` and the `Initialize` instruction are extended with a `twap_oracle_program_id` that the instruction reads.

### DIFF
--- a/programs/amm/src/create_price_observations.rs
+++ b/programs/amm/src/create_price_observations.rs
@@ -1,0 +1,434 @@
+use amm_core::{
+    compute_config_pda, compute_pool_pda, compute_pool_pda_seed, AmmConfig, PoolDefinition,
+};
+use clock_core::CLOCK_01_PROGRAM_ACCOUNT_ID;
+use nssa_core::{
+    account::{Account, AccountWithMetadata},
+    program::{AccountPostState, ChainedCall, ProgramId},
+};
+use twap_oracle_core::{
+    compute_current_tick_account_pda, compute_price_observations_pda, CurrentTickAccount,
+};
+
+/// Creates a TWAP price-observations account for `pool` over a time window, on behalf of the AMM.
+///
+/// The pool acts as the price source: the AMM authorizes it via its pool PDA seed so the oracle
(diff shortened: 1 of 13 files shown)


gravityblast (approved):

programs/amm/src/create_price_observations.rs
```diff
@@ -0,0 +1,434 @@
...
+            pool_def.definition_token_a_id,
+            pool_def.definition_token_b_id,
+        ),
+        "Create price observations: Pool Account ID does not match PDA"
+    );
+
+    // The initial tick comes from the pool's authoritative CurrentTickAccount, not from the
+    // caller. Verifying its PDA ties it to this exact pool, so the seed tick cannot be forged.
+    assert_eq!(
+        current_tick_account.account_id,
```
3esmit, line 84: This is not ready yet, I assume this PR don't expect full functionality? Because it seems that no AMM instruction currently  creates or update the pool current-tick account, so creating an observation would fail here or use a stale value.  

The tests implemented on the PR bypass this by using `force_insert_account`.

If you want to implement it on this PR, than on AMM side you would want the tick lifecycle implemented there, so it can properly call create price observations.

3esmit (commented):
Approved with comments, confim this is the scope of the implementation.

programs/amm/src/create_price_observations.rs
0x-r4bbit: The creation of the tick account happens in #149

PR #149, opened 2026-06-18
feat(amm): bootstrap pool TWAP current-tick account at pool creation

0x-r4bbit:
Extend new_definition to also create the pool's TWAP current-tick account via a chained CreateCurrentTickAccount, so a pool and its price feed are born together. The opening tick is derived on-chain from the pool's own reserves (reserve_b / reserve_a as Q64.64), not caller-supplied, so it cannot be forged. The pool is passed in its post-claim state and authorized as the price source via its pool PDA seed.

Add spot_price_q64_64 to amm_core (not the oracle): the reserves -> price mapping is the price source's concern; the oracle only converts price to a tick.


programs/amm/src/new_definition.rs
```diff
@@ -202,11 +219,41 @@ pub fn new_definition(
...
+        account_id: pool.account_id,
+    };
+    let call_create_current_tick = ChainedCall::new(
+        twap_oracle_program_id,
+        vec![
+            current_tick_account.clone(),
+            pool_price_source,
+            clock.clone(),
+        ],
+        &twap_oracle_core::Instruction::CreateCurrentTickAccount { initial_price },
```
3esmit, line 244: This creates the pool `CurrentTickAccount` with the opening price, but there is nowhere a `UpdateCurrentTick` call. 

I assume this is due the scope of PR is to only bootstrap pool.

3esmit (commented):
Approved with comments.

programs/amm/src/new_definition.rs
0x-r4bbit: Correct! #150 addresses updating the tick

PR #150, opened 2026-06-18
feat(amm): refresh TWAP current tick on add/remove liquidity

0x-r4bbit:
Extend add_liquidity and remove_liquidity to chain UpdateCurrentTick to the oracle, matching swaps and sync_reserves. Each takes the current-tick and clock accounts, reads the TWAP program ID from config, validates the clock and current-tick PDA, and after computing the post-op pool chains an UpdateCurrentTick carrying the post-op spot price with the pool as the authorized price source.

Every reserve-mutating instruction now keeps the current tick in sync, so current_tick == tick(reserves) holds after each operation. Proportional add/remove preserve the price, so the tick is unchanged for them, but the refresh still runs and lands on the correct value.

### DIFF
--- a/artifacts/amm-idl.json
+++ b/artifacts/amm-idl.json
@@ -244,6 +244,18 @@
           "writable": false,
           "signer": false,
           "init": false
+        },
+        {
+          "name": "current_tick_account",
+          "writable": false,
+          "signer": false,
+          "init": false
+        },
+        {
+          "name": "clock",
+          "writable": false,
+          "signer": false,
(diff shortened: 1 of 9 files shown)


3esmit (approved):
Approved with suggestions: 
- update `amm/core/src/lib.rs` comments that describe the changed required accounts.
- add one successful exact-output integration test that executes the transaction and checks the stored `CurrentTickAccount` changed to the post-swap spot price.

PR #192, opened 2026-06-22
feat(stablecoin): add fixed-point math utilities

gravityblast:
closes #157

### DIFF
--- a/programs/stablecoin/core/Cargo.toml
+++ b/programs/stablecoin/core/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "v0.2.0-rc3", features = ["host"] }
 borsh = { version = "1.5", features = ["derive"] }
+primitive-types = { version = "0.13", default-features = false }
 serde = { version = "1.0", features = ["derive"] }
 twap_oracle_core = { path = "../../twap_oracle/core" }
 risc0-zkvm = { version = "=3.0.5", default-features = false }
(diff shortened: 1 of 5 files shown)


programs/stablecoin/core/src/math.rs
```diff
@@ -0,0 +1,214 @@
...
+/// # Edge cases
+/// - `milliseconds_elapsed == 0` returns `FIXED_POINT_ONE` (identity).
+/// - `per_millisecond_rate == FIXED_POINT_ONE` returns `FIXED_POINT_ONE` regardless of
+///   `milliseconds_elapsed`.
+///
+/// # Overflow
+/// NOT self-bounding. For any `per_millisecond_rate > FIXED_POINT_ONE` this
+/// eventually overflows `u128` as `milliseconds_elapsed` grows — the §8 rate
+/// bound alone does not prevent it. Callers MUST clamp the elapsed window to
+/// `MAXIMUM_COMPOUNDING_WINDOW_MILLISECONDS` (spec §5.3) before calling.
```
0x-r4bbit: Not an issue atm but we're referring to a constant that doesn't exist yet

0x-r4bbit: Also, I'm wondering if we should leave 

> (spec §5.3)

^ stuff like this out of the code. What do you think?

programs/stablecoin/core/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 [dependencies]
 nssa_core = { git = "https://github.com/logos-blockchain/logos-execution-zone.git", tag = "v0.2.0-rc3", features = ["host"] }
 borsh = { version = "1.5", features = ["derive"] }
+primitive-types = { version = "0.13", default-features = false }
```
0x-r4bbit: We have `alloy-primitives` in latest `main`, can we consolidate those?

0x-r4bbit (approved):
Approving, but I think we can avoid having two primitives libraries.
We need alloy-primitives for twap for the uniswapv3 math.

If it can also serve the implementation here, I'd suggest we drop `primitive-types`.

PR #197, opened 2026-06-24
feat(amm): wire the AMM app to the LEZ wallet module

0x-r4bbit:
Turns the dummy-data AMM UI into a real client of the on-chain LEZ wallet. Adds a hand-written ui_qml C++ backend (src/AmmUi*) over the core logos_execution_zone module: create/open a local wallet, create and list public/private accounts, and a navbar Connect / Connected + account-selector
+ Disconnect flow. Onboarding is password-only (no path picking) with the canonical LEZ wallet at ~/.lee/wallet (override: LEE_WALLET_HOME_DIR); standalone shares that keystore with other LEZ wallet clients, Basecamp shares accounts via adopt-on-start.

Requires Nix with flakes. macOS also needs Xcode + the Metal Toolchain (RISC0 builds GPU prover kernels):

    xcodebuild -downloadComponent MetalToolchain

If the build still reports "missing Metal Toolchain", reboot (Xcode 26 mounts the toolchain per-user; a reboot makes it visible to nix's build users). Builds also require nix `sandbox = false` (default on macOS). Linux needs none of the Metal steps.

The pinned core module still references a pre-fix LEZ rev, so the LEZ input is repointed to a build that includes the macOS Metal fix (github:<LEZ-with-fix>), already wired in apps/amm/flake.nix — plain `nix run .` works:

    cd apps/amm
    nix run .

- The wallet password is currently a no-op upstream (storage.rs: "TODO: use password for storage encryption"); storage.json is plaintext. So Disconnect is a UI-level lock and reconnect does not (cannot yet) re-prompt for it.
- wallet-ffi requires explicit config/storage paths; a *_default() FFI would let the app drop its path handling.
- Bundled network config: connects to whatever WalletConfig::default() points at; real testnet endpoints still TBD.


apps/amm/result
3esmit, line 1: This file should not be commited

apps/amm/src/AmmUiBackend.cpp
```diff
@@ -0,0 +1,381 @@
...
+{
+    // Respect an explicit user disconnect: stay locked, show "Connect".
+    if (QSettings(SETTINGS_ORG, SETTINGS_APP).value(DISCONNECTED_KEY, false).toBool())
+        return;
+
+    // In Basecamp the logos_execution_zone module is a single shared instance,
+    // so the wallet may already be open (e.g. opened by the dedicated wallet
+    // app). Adopt that wallet instead of fighting over it: if list_accounts()
+    // returns anything, treat the wallet as open and just mirror its state.
+    const QJsonArray existing = QJsonArray::fromVariantList(m_logos->logos_execution_zone.list_accounts());
```
3esmit: This consider that `list_accounts() == []` is "wallet not open", but if shared wallet can have zero accounts and already be open. In that case this falls through to `open(cfg, stg)`, and that would rejects a second open when `walletHandle` already exists, leaving AMM disconnected with no way to Add Account.

apps/amm/src/AmmUiBackend.cpp
```diff
@@ -0,0 +1,1676 @@
...
+}
+
+bool AmmUiBackend::createNew(QString configPath, QString storagePath, QString password)
+{
+    const QString localConfig = toLocalPath(configPath);
+    const QString localStorage = toLocalPath(storagePath);
+    // create_new now returns the new wallet's BIP39 mnemonic (empty on failure),
+    // not an int status code. TODO: surface this seed phrase to the user for
+    // backup — it is currently discarded, so the wallet can't be recovered.
+    const QString mnemonic = m_logos->logos_execution_zone.create_new(localConfig, localStorage, password);
```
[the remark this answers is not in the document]
3esmit: Second this

3esmit (dismissed):

PR #199, opened 2026-06-26
feat(stablecoin): add ProtocolParameters account type

gravityblast:
closes #158


3esmit (approved):
Everything good. Just a heads up that future PRs on initialization and setter must use the singleton PDA helper consistently when deriving the singleton account, and they should enforce the numeric bounds documented in `docs/stablecoin/README.md` section 8.

PR #206, opened 2026-06-28
Refactor/lez v0.2.0 rc6 compat

0x-r4bbit:
## Summary

- Update the workspace to the LEZ `v0.2.0-rc6` dependency set, including the `lee` / `lee_core` package rename while preserving the existing `nssa` / `nssa_core` dependency names.
- Pin the SPEL framework crates to an immutable revision and refresh the affected Cargo lockfiles.
- Harden AMM arithmetic by widening intermediate products to `U256` for liquidity creation, add/remove liquidity, and exact input/output swap calculations.
- Add regression coverage for large 18-decimal pool amounts, overflow-boundary liquidity math, and exact-output slippage behavior.
- Add PDA helper examples for AMM, ATA, and TWAP oracle flows, refresh AMM/TWAP IDL artifacts, and add a testnet runbook for build, deploy, pool setup, swaps, and TWAP publishing.

## Compatibility Notes

- Integration and benchmark tests are updated for the rc6 API surface, explicit `clock_core` usage, and `LeeError` result types.
- Guest manifests and lockfiles are regenerated against the rc6 graph.
- TWAP oracle instruction ordering is adjusted so the generated IDL stays in sync with the guest entrypoint.

### DIFF
--- a/docs/testnet-runbook.md
+++ b/docs/testnet-runbook.md
@@ -1,0 +1,490 @@
+# AMM on testnet — spel runbook
+
+End-to-end steps to deploy the programs, initialize the AMM, and create a pool using
+`spel`, in the order they must happen. Follow top to bottom; nothing here can be skipped.
+
+> **Golden rule:** every time you **recompile** a guest, its **ProgramId changes**, and
+> **every PDA derived from that ProgramId changes too** (config, pool, vaults, LP def, lp
+> lock, current tick). If you rebuild the AMM, you must recompute all AMM PDAs and
+> re-run `initialize`. Treat ProgramIds and PDAs as deployment-specific — re-derive them,
+> never reuse old values.
+
+---
+
+## Contents
+
+- [0. Prerequisites](#0-prerequisites)
+  - [Argument formats](#argument-formats-used-throughout)
+- [1. Build & deploy the programs](#1-build--deploy-the-programs)
+  - [Record the ProgramIds](#record-the-programids)
+- [2. Wallet CLI basics](#2-wallet-cli-basics)
+- [3. Create two token definitions](#3-create-two-token-definitions)
+  - [(Optional) Transfer tokens](#optional-transfer-tokens)
+- [4. Derive the AMM PDAs](#4-derive-the-amm-pdas)
+  - [All PDA helpers](#all-pda-helpers)
+- [5. Initialize the AMM](#5-initialize-the-amm)
+- [6. Create a pool (new-definition)](#6-create-a-pool-new-definition)
+- [7. Verify](#7-verify)
+- [8. (Optional) Create a TWAP price-observations account](#8-optional-create-a-twap-price-observations-account)
+- [9. Swap](#9-swap)
+- [10. Record a tick](#10-record-a-tick)
+- [11. Create the oracle price account](#11-create-the-oracle-price-account)
+- [12. Publish a price](#12-publish-a-price)
+- [Gotchas](#gotchas-we-hit-and-how-to-avoid-them)
+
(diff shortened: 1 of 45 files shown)


programs/amm/methods/guest/src/bin/amm.rs
```diff
@@ -312,9 +312,9 @@ mod amm {
         vault_a: AccountWithMetadata,
         #[account(mut)]
         vault_b: AccountWithMetadata,
-        #[account(mut)]
+        #[account(mut, signer)]
         user_holding_a: AccountWithMetadata,
-        #[account(mut)]
+        #[account(mut, signer)]
```
3esmit, line 317: So for a trade to happen, we need a signature from both sides? How the user_holding_a and user_holding_b syncronize this in real world?

programs/amm/methods/guest/src/bin/amm.rs
```diff
@@ -163,7 +163,7 @@ mod amm {
         user_holding_a: AccountWithMetadata,
         #[account(mut, signer)]
         user_holding_b: AccountWithMetadata,
-        #[account(mut)]
+        #[account(mut, signer)]
```
3esmit, line 166: With this, the user_holding_b must sign even when the LP holding already exists. 

Runtime check in `new_definition.rs` only requires authorization for a default/fresh LP target. 

If this is correct, update the conditional checks and the comment/docs;

programs/amm/methods/guest/src/bin/amm.rs
0x-r4bbit: Actually you only need the signature for the account from which you send tokens **in**
But we don't know which one will be used as token in, so the quick fix is to supply sigantures for both then you can choose freely.

Definitlly room for improvement.

programs/amm/methods/guest/src/bin/amm.rs
0x-r4bbit: There's actually no case where the LP holding can already exist. It has to be uninitialized.
THe condition can be updated to reflect that

programs/amm/methods/guest/src/bin/amm.rs
3esmit: Wouldnt SPEL force requirement on both sides to sign it? Perhaps SPEL needs a `?signer` or `maybe-signer`  somethign like that;

0x-r4bbit: Yes, so in this case, no matter which direction you're trading, you'll have to supply sigantures for both token accounts (even if just one is receiving).

Some we can fix later

docs/testnet-runbook.md
```diff
@@ -0,0 +1,448 @@
...
+
+- `<DEF_A>` becomes the token-definition account; `<HOLDING_A>` receives `total_supply`.
+- Repeat for token B (`<DEF_B>`, `<HOLDING_B>`).
+- These `<DEF_*>` ids are what feed all the AMM pool PDAs.
+
+Inspect to confirm / find a holding's definition id:
+
+```bash
+spel --idl artifacts/token-idl.json -- inspect <HOLDING_A> --type TokenHolding
+spel --idl artifacts/token-idl.json -- inspect <DEF_A>     --type TokenDefinition
```
gravityblast:
```suggestion
spel --idl artifacts/token-idl.json inspect <HOLDING_A> --type TokenHolding
spel --idl artifacts/token-idl.json inspect <DEF_A>     --type TokenDefinition
```

3esmit (approved):

PR #215, opened 2026-07-07
test(privacy): add privacy-preserving test coverage for LEZ programs

jonesmarvin8:
## 🎯 Purpose

Explore privacy claims of LEZ against LEZ programs. Add tests to demonstrate these capabilities, and explore where the features are not expressible.

## ⚙️ Approach

For each LEZ program (token, amm, stablecoin, ata): we add privacy tests to explore functionality against REGULAR private accounts (initialized and uninitialized accounts using `nsk`), EXIST (private account initialization using `npk`), and GROUP (private account shared among a group).

Private PDAs were not testable with these programs due to the `AccountId` formulas differing from the public variant's formula.

`docs/findings.md` gives an overview of the findings and tests.
`docs/privacy-test-matrix.md` is an AI (Claude) generated document that provides more context generated (has not been edited).

## 🧪 How to Test

`RISC0_DEV_MODE=1 cargo test`

## 🔗 Dependencies

N/A

## 🔜 Future Work

`findings.md` outlines minor issues/fixes that can be applied to LEZ, and a bug in `spel-framework`. Git issues for each of these will be opened.

## 📋 PR Completion Checklist

*Mark only completed items. A complete PR should have all boxes ticked.*

- [X] Complete PR description
- [X] Implement the core functionality
- [X] Add/update tests
- [X] Add/update documentation and inline comments


3esmit (approved):

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+LEE's private state supports (regular) accounts, PDAs and group owned accounts.
+
+## Overview of (regular) private accounts
+
+### Private account initialization
+Regular private accounts can be initialized with or without knowledge of the account's nullifier secret key `nsk`. This results in two initialization "types": `PrivateUnauthorized` and `PrivateAuthorizedInit`.
+
+- `PrivateUnauthorized`
+
+    A special case for private accounts initialization that uses only public keys `npk` and `vpk`. Example: Alice can use Bob's keys (`npk`, `vpk`) and an `identifier` to send Bob a private transaction. Since Alice does not know the corresponding `nsk`, she is spend the resulting private account. E.g., Alice cannot authorize the transaction.
```
0x-r4bbit: >  she is spend the resulting private account

This doesn't make a lot of sense to me. Would you mind clarifying?

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+|type | authorized | who can use |
+|----|----|----|
+| `PrivateUnauthorized`| &#10060; | anyone |
+| `PrivateAuthorizedInit` | &#9989; | owner |
+| `PrivateAuthorizedUpdate` | &#9989; | owner |
+
+Only the account owner can (1) update their initialized account, and (2) use functions that require authorization with their account.
+
+### Remark
+- `PrivateUnauthorized` initialization is used for account initialization. `is_authorized = false` is a protection that does not seem crucial. Artifically, blocks some functions.
```
0x-r4bbit: > `is_authorized = false` is a protection that does not seem crucial. Artifically, blocks some functions.

Can you clarify what exactly this is trying to say? `is_authorized` is actually quite crucial for atm. But maybe you're referring to something else?

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+handing over their actual secret key. The mechanism is a **Group Master Secret (GMS)**,
+distributed via a real seal/unseal handshake (ML-KEM-768), not key reuse:
+
+1. Alice creates a `GroupKeyHolder` and derives the shared account's keys (`nsk`, `vsk`)
+   from it.
+2. Alice **seals** the GMS against Bob's sealing public key and hands over only the sealed bytes.
+3. Bob **unseals** it with his own sealing secret key, then
+   independently re-derives the account's keys from the same seed.
+
+This ensures that any member of the group can execute programs on shared accounts using either `PrivateAuthorizedInit` or `PrivateAuthorizedUpdate`. From a program's perspective, shared accounts should behave the same as regular public accounts.
```
0x-r4bbit: > From a program's perspective, shared accounts should behave the same as regular public accounts

Not regular private accounts?

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+| SwapExactInput | `amm_swap_a_to_b_private_authorized_init_destination_is_not_expressible` | REGULAR, CHAIN | Swap paying out to a brand-new `PrivateAuthorizedInit` destination (owner self-initializes with its own `nsk`) | ❌ (confirmed not-expressible — same "destination must already exist" precondition as `RemoveLiquidity`) |
+| NewDefinition | `amm_new_definition_private_initial_lp_holder` | REGULAR | Pool creation with a private `PrivateAuthorizedInit` initial LP holder | ✅ |
+| NewDefinition | `amm_new_definition_private_unauthorized_lp_holder_is_not_expressible` | EXIST, REGULAR | Pool creation with a `PrivateUnauthorized` initial LP holder (`npk` only, no `nsk`) | ❌ (confirmed not-expressible — guest ABI requires `user_holding_lp` to be a signer, which `PrivateUnauthorized` can never satisfy; same shape as the `Swap` `PrivateUnauthorized` finding above) |
+
```
0x-r4bbit, line 94: The `confirmed not-expressible` cases are okay right? Can you confirm?

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+### Remarks
+- `Swap` and `Remove` rejects any uninitialized destination account; this is a AMM design choice, and not Token program requirement.
+- AMM tests were initially blocked by a bug.
```
0x-r4bbit: What bug?

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+
+
+### Remarks
+- `Initialization` is not possible for `PrivateUnauthorized` accounts due to `is_authorized = false`.
+- New token definition is not permitted for `PrivateUnauthorized` as Token holding due to `is_authorized = false`.E.g., both Token Definition and Token Holding for a new Token must be from an authorized account.
+
+# Conclusions
+
+Privacy coverage for LEZ program tests is greatly improved from the added tests. Though, there are a few noticable gaps:
+- `PrivateUnauthorized` accounts can be blocked by programs with a check `is_authorized = true`. However, this issue can be avoided by defining `is_authorized = true` for account initialization with `PrivateUnauthorized` (e.g., no knowledge of `npk`). Account initialization cannot be used to maliciously alter a pre-existing account, and thus `is_authorized = true` would not offer any malicious path forward for the third-party initializing the account.
```
0x-r4bbit: Can you elaborate what you're suggesting here? `is_authorized` is typically set by the LEZ environment. We don't control that unless its a PDA.

docs/findings.md
```diff
@@ -0,0 +1,165 @@
...
+- Privacy transactions have issues with chain calls in which multiple calls affect the same private account. This issue can be mitigated by adopting account diff paradigm instead of the current "account state replacement" that we currently use.
```
0x-r4bbit: Any example of that? Do we have cases where this applies in our programs?

0x-r4bbit (commented):
@jonesmarvin8 Thank you for this!

I've left some comments. Maybe you can clarify those.
To me it's a little bit unclear if there's any action needed on our side. 

Some comments in the docs suggest we need to adjust the programs, but it's not super explicit.

Can you confirm this?

docs/findings.md
jonesmarvin8: Disclaimer: the bug was found while working with `lez-programs` but it's a bug in LEE itself (opened a PR to patch). I just wanted to make that clear.

For public transactions, we require that `inputs accounts == output accounts`; these sets are defined within the function's execution. However, there's nothing that explicitly checked to make sure the returned input accounts match with the function's parameters. E.g., we can quietly drop (or add) an input account to what the program returns.

This is blocked explicitly by privacy transactions. So, while I was writing tests...I kept getting this issue seemingly "blocking" me.

docs/findings.md
jonesmarvin8: Absolutely, I'm suggesting that at the LEE protocol level...we set `is_authorized = true` for `PrivateUnauthorized` accounts. These accounts are initialized by a third-party for the user. So, (imo) there's no good reason for the protocol level to block program behavior. Especially since the third-party can only initialize an account, they cannot update a pre-existing account.

docs/findings.md
jonesmarvin8: `not-expressible` are cases that would be ideal to allow for. But, with the current architecture are blocked.

For `is_authorized = false` for `Private::Unauthorized` (as mentioned in the next comment), this is a LEE protocol fix.

For `amm_swap_a_to_b_private_authorized_init_destination_is_not_expressible`, this can be fixed by removing the constraint that destination account is not default. This is a program fix.

docs/findings.md
jonesmarvin8: I mistyped. Thanks!

docs/findings.md
jonesmarvin8: Yes. I'll provide additional context in the document as well.

`is_authorized` is crucial for public accounts. Because public account ids can be "accessed" and used by anyone in a transaction. Private accounts cannot.

A private account can only be updated by the account owner (using the `nsk`). So, this specific change that I recommend is a protocol level change only for the branching logic for `Private::Unauthorized`. This is the case in which a third-party uses a published `npk`, `vpk` and some `identifier` (of their choosing) to initialize a new account for the account owner. Since this is a newly initialized account (and cannot be updated without the `nsk`), I see no reason why at the protocol level we cannot just set `is_authorized = true`. The ramifications of allowing `is_authorized = true` only extends to a newly initialized account which a user can choose to use or not use, but cannot impact any of their pre-existing accounts.

docs/findings.md
jonesmarvin8: Yes....apparently my knowledge of written language gave up when I was writing...and proofreading this section. My apologies.

It should have been "Since Alice does not know the corresponding `nsk`, she is unable to spend the resulting private account."

docs/findings.md
jonesmarvin8: Covering both issues I see listed:
- `PrivateUnauthorized` several functions have this issue, but a non issue. ([PR 621](https://github.com/logos-blockchain/logos-execution-zone/pull/621) handles this issue, and will be merged today. I'll update the tests accordingly).

- Privacy transactions have issues with chain calls. This was actually an interesting issue that appeared in a single function (Stablecoin's `OpenPosition`). What happens is initialization and token transfer occurs within the same chain call. The public version allows for this but it is blocked by the privacy version (due to the privacy preserving circuit). As mentioned, we have a proposed revision to the privacy preserving circuit that would fix this issue altogether. 

IMO, I don't think devs should be (artificially) constrained by LEZ for "good" coding practices. E.g., initialize Token account and transfer to the same account seems to be reasonable behavior for a dev to try; even if strange to me since token transfer initializes. Instead, I would rather a protocol fix that allows this reasonable behavior to work in both public and privacy transfers.

jonesmarvin8: PR 621 is merged to `logos-execution-zone` 's `dev` branch; the `dev` branch is the active branch for new developments and work. `main` is the current build of logos.

docs/findings.md
jonesmarvin8: I've added more details to the document for this.

docs/findings.md
jonesmarvin8: I made the tables clearer.

0x-r4bbit:
Closing this in favour of #350 (no offense @jonesmarvin8 but since this PR is coming from your fork I had to open a new one to update it).

PR #230, opened 2026-07-15
feat(apps/token): add token definition prototype

0x-r4bbit:
## Summary

This PR evolves the Token app from a shared-wallet scaffold into an interactive, local-only Token Program prototype.

It gives operators a concrete workbench for preparing fungible and non-fungible token definitions before a production transaction API is connected.

## Screenshots

image
image
image
image

image

image


apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+            }
+            refreshSequencerAddr();
+            refreshAccounts();
+        }
+        return true;
+    }
+
+    void copyToClipboard(QString text) override
+    {
+        if (QGuiApplication::clipboard())
```
3esmit, line 298: The shared QML already documents that calling `QGuiApplication::clipboard()` in the headless module host tears the backend down, and it provides a GUI-side `TextEdit.copy()` helper for address/account copies. The mnemonic path still calls this backend slot (`CreateWalletDialog.onCopyRequested` -> `root.backend.copyToClipboard`), so pressing “Copy to clipboard” on the backup page can drop the new wallet connection. Route this remaining copy through the existing QML helper and remove the headless clipboard access.

apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+            QSettings(SETTINGS_ORG, m_settingsApp).setValue(DISCONNECTED_KEY, false);
+            return true;
+        }
+
+        const QString cfg = this->configPath().isEmpty() ? defaultConfigPath() : this->configPath();
+        const QString stg = this->storagePath().isEmpty() ? defaultStoragePath() : this->storagePath();
+        if (!QFileInfo::exists(stg))
+            return false;
+
+        const int err = m_logos->logos_execution_zone.open(cfg, stg);
```
3esmit, line 227: require `logos::CallError::ok()` as well as return code zero here. The generated `open()` wrapper returns `QVariant::toInt()`; when the remote call fails, the invalid variant converts to `0`, which is also `WALLET_FFI_SUCCESS`. This path then marks the wallet open and returns `true` even though no wallet was opened. The same check is needed for the startup and sequencer-reopen calls at lines 360 and 285.

apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+        const QJsonArray arr = QJsonArray::fromVariantList(m_logos->logos_execution_zone.list_accounts());
+        m_accountModel->replaceFromJsonArray(arr);
+        refreshBalances();
+    }
+
+    void refreshBalances() override
+    {
+        refreshBlockHeights();
+        if (this->currentBlockHeight() > 0)
+            m_logos->logos_execution_zone.sync_to_block(static_cast<quint64>(this->currentBlockHeight()));
```
3esmit, line 160: `sync_to_block()` returns an FFI status, but it is discarded. If the sequencer sync returns nonzero, this method still reads balances from the pre-sync wallet state, saves that state, and completes the refresh as if it succeeded, so both apps can display stale balances without an error. Check the status (and `CallError`) and abort or propagate failure before reading and persisting balances.

apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+
+        const QString cfg = this->configPath().isEmpty() ? defaultConfigPath() : this->configPath();
+
+        // Preserve the other config fields (poll timeouts, retries) — only swap
+        // the endpoint. The wallet reads this file on open via
+        // from_path_or_initialize_default.
+        QJsonObject obj;
+        QFile in(cfg);
+        if (in.open(QIODevice::ReadOnly)) {
+            obj = QJsonDocument::fromJson(in.readAll()).object();
```
3esmit, line 269: This can silently destroy the existing wallet config. A truncated config or valid JSON array makes `.object()` return an empty object, after which the code truncates the original and writes only `sequencer_addr`. Even with valid JSON, `QFile::write()` can fail or short-write after `open()` succeeds, but its result is ignored and a closed-wallet path still returns `true`. Validate `QJsonParseError`/`isObject()`, then write through `QSaveFile` and require a successful full write plus `commit()` before replacing the original.

apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+    void checkReachability()
+    {
+        const QString addr = this->sequencerAddr();
+        if (addr.isEmpty())
+            return;
+
+        QNetworkRequest req{QUrl(addr)};
+        req.setTransferTimeout(4000);
+        QNetworkReply* reply = m_net->get(req);
+        QObject::connect(reply, &QNetworkReply::finished, this, [this, reply]() {
```
3esmit, line 416: This callback can overwrite status for a newer sequencer address. If a periodic probe for unreachable A is pending, the user changes to reachable B, and B finishes first, the later timeout from A sets `sequencerReachable=false` again even though `sequencerAddr` is B. Capture the probed address or a generation and update only if it still matches the current endpoint, or abort superseded replies.

apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+    {
+        saveWallet();
+        delete m_logos;
+    }
+
+    AccountModel* accountModel() const { return m_accountModel; }
+
+    // ── .rep slot overrides ──────────────────────────────────────────────────
+
+    QString createAccountPublic() override
```
3esmit, line 133: Inside Basecamp both backends adopt one shared `logos_execution_zone` wallet, but each keeps its own `AccountModel` snapshot. Creating an account here refreshes only the initiating app's model; the other app receives no signal, subscription, or periodic account refresh and keeps the old account list indefinitely. Publish a shared wallet-change notification/central model, or otherwise refresh every consumer when the shared account set mutates.

apps/common/wallet-ui/src/WalletBackendLogic.h
```diff
@@ -0,0 +1,453 @@
...
+            if (this->sequencerReachable() != reachable)
+                this->setSequencerReachable(reachable);
+            reply->deleteLater();
+        });
+    }
+
+    void saveWallet()
+    {
+        if (this->isWalletOpen())
+            m_logos->logos_execution_zone.save();
```
3esmit, line 432: `save()` returns both an FFI status and `CallError`, but both are discarded. For example, after `createAccountPublic()` mutates the wallet, a read-only/full storage path makes persistence fail while the slot still returns the new account ID and QML reports success; `disconnectWallet()` likewise clears the model and marks the app disconnected after a failed save. Check and propagate the save result, and do not complete disconnect or report a durable mutation until persistence succeeds.

3esmit (dismissed):

PR #232, opened 2026-07-22
feat(amm): headless amm_module core Logos module; flip UI to consume it

0x-r4bbit:
Introduce modules/amm — the AMM business logic (resolvePool, swapExactInput, tokenList) as a universal core Logos module that wraps amm_client_ffi and depends on the logos_execution_zone wallet module. The same surface is now consumed by the QML UI (via modules().amm_module) and headlessly (logoscore call amm_module ...).

- apps/amm: drop the direct amm_client_ffi link; depend on and delegate to amm_module (metadata dependencies, CMakeLists, AmmUiBackend). Wallet-session UI logic stays in the backend.
- flake.nix: build amm_module via mkLogosModule and inject it into the UI builder's flakeInputs so the dependency resolves.
- amounts/deadline declared nlohmann::json so the generated dispatch accepts a JSON number (bare small ints on the CLI) or a string (exact u128 from the UI, or a quote-wrapped big value on the CLI); JSON floats are rejected rather than submit a silently-rounded amount.
- AMM_DEBUG-gated tracing for the swap path.
- modules/amm/README.md: headless prerequisites + full logoscore recipe.


apps/amm/src/AmmUiBackend.cpp
```diff
@@ -177,189 +197,37 @@ void AmmUiBackend::syncWalletState()
...
 }
 
 QString AmmUiBackend::swapExactInput(QString defAHex, QString defBHex, QString userInputHoldingHex,
                                       QString userOutputHoldingHex, QString amountInDecimal,
                                       QString minOutDecimal, QString deadlineDecimal)
 {
-    const QString txHash = m_swap->swap(defAHex, defBHex, userInputHoldingHex, userOutputHoldingHex,
-                                        amountInDecimal, minOutDecimal, deadlineDecimal,
-                                        networkSnapshot(), isWalletOpen());
+    const QString txHash = m_logos->amm_module.swapExactInput(
```
3esmit, line 226: This forwarding path drops the wallet-open guard that `SwapRuntime::swap()` had. `disconnectWallet()` only marks this app disconnected; it intentionally leaves the shared `logos_execution_zone` wallet handle open, and `SwapCard::canSubmit` does not check `isWalletOpen`. After opening then disconnecting ~ or while another app keeps the shared wallet open ~ confirming a swap still reaches `send_generic_public_transaction` and can sign/submit while this UI shows Connect. 
fix: restore `if (!isWalletOpen()) return {};` before forwarding, or pass and enforce this app's wallet-open state in `amm_module`.

modules/amm/src/amm_module_impl.cpp
```diff
@@ -0,0 +1,742 @@
...
+nlohmann::json AmmModuleImpl::readConfig(const Network& net) {
+    const FfiResult configResult =
+        call(amm_config_id, json{{"ammProgramId", net.amm_program_id}});
+    if (!configResult.ok) return json();  // null: config_id op failed
+    return readPublicAccount(jStr(configResult.value, "configId"));
+}
+
+LogosMap AmmModuleImpl::resolvePool(const std::string& def_a_hex,
+                                    const std::string& def_b_hex) {
+    const json absent = {{"exists", false}};
```
3esmit: Every early failure now becomes only `{"exists": false}`. For example, with `AMM_PROGRAM_BIN` unset, `network().status` is `config_missing`, but line 410 discards it even though this header and README promise an `error` field (`no_program_bin` in that case). The same shape also hides config-ID, account-read, and FFI failures as an ordinary missing pool, so `SwapCard` cannot surface its existing `pool.error` state. 
fix: return a stable error code at each failure boundary, or narrow the public contract and consuming UI deliberately.

modules/amm/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+6. **A tokens config** for `TOKENS_CONFIG` — a JSON array of
+   `{ symbol, name, definitionId, holding, decimals }` (e.g. the repo's
+   `amm-tokens.json`).
+
+7. **A wallet** at `~/.lee/wallet` (`wallet_config.json` with `sequencer_addr`
+   pointing at your sequencer, plus `storage.json` with your accounts), and a
+   **running sequencer** with the AMM initialized and a pool holding liquidity.
+   Swaps also need the wallet **open + synced** (see below); reads
+   (`resolvePool`/`tokenList`) do not.
```
3esmit: `resolvePool` cannot currently run before the wallet is open. Its path is `resolvePool` -> `readConfig`/`readPublicAccount` -> `logos_execution_zone.get_account_public`; the pinned wallet module passes its `walletHandle` to `wallet_ffi_get_account_public`, and that handle is null until `open` or `create_new`. Consequently the headless recipe at lines 200-205 reports an absent pool even when one exists. 

Fix: require/open the wallet before this read, or provide a sequencer-read path that does not depend on a wallet handle.

modules/amm/src/amm_module_impl.cpp
```diff
@@ -0,0 +1,742 @@
...
+    // word (matches `spel program-id` and the on-chain *_program_id fields).
+    const FfiResult r = call(amm_program_id, json{{"elf", toHex(elf.data(), elf.size())}});
+    if (!r.ok) {
+        AMM_TRACE("ammProgramId: amm_program_id op failed");
+        return {};
+    }
+    return jStr(r.value, "programId");
+}
+
+AmmModuleImpl::Network AmmModuleImpl::network() {
```
3esmit, line 308: `network()` now rebuilds invariant process configuration on every `resolvePool`, swap, context, quote, and submit call: it rereads the whole AMM ELF, hex-expands and hashes it through `amm_program_id`, rereads `TOKENS_CONFIG`, and may make synchronous `account_id_from_base58` calls for every token. The generated Qt wrapper is synchronous, and `publishNetworkContext()` invokes this path on every wallet-state change. The replaced backend deliberately cached these process-lifetime inputs because this is a hot/reply path. 

fix: restore one-time module caching, with explicit invalidation only if runtime env reload is intended.

modules/amm/src/amm_module_impl.cpp
```diff
@@ -0,0 +1,742 @@
...
+    result["account"] = {
+        {"program_owner", owner},
+        {"balance", balance},
+        {"nonce", nonce},
+        {"data", data},
+    };
+    return result;
+}
+
+nlohmann::json AmmModuleImpl::walletAccountReads(bool wallet_open, bool /*refresh*/) {
```
3esmit: `refresh` is unused, so `newPositionContext(..., false)` and `newPositionContext(..., true)` both perform the same full `list_accounts` plus `get_account_public` reads. This breaks the generated API contract that `true` forces a fresh read rather than a cached one, and leaves the QML's deliberate refresh flag meaningless.

modules/amm/src/amm_module_impl.cpp
```diff
@@ -0,0 +1,742 @@
...
+        const std::string definition_id = normalizeAccountId(jStr(entry, "definitionId"));
+        const std::string holding = normalizeAccountId(jStr(entry, "holding"));
+        if (definition_id.empty() || holding.empty()) continue;
+
+        json token;
+        token["symbol"] = jStr(entry, "symbol");
+        token["name"] = jStr(entry, "name");
+        token["definitionId"] = definition_id;
+        token["holding"] = holding;
+        token["decimals"] = entry.value("decimals", 0);
```
3esmit: `entry.value("decimals", 0)` defaults only a missing key; a present non-integer value such as `"decimals": "18"` throws `json.exception.type_error.302`. The generated dispatch converts that exception to `dispatch_failed`, and the Qt caller then sees an empty list, so one malformed external config entry removes every token.

3esmit (dismissed):

PR #236, opened 2026-07-27
refactor(amm): consolidate the two client FFIs into one JSON crate

0x-r4bbit:
The AMM host FFI was split across two crates with two ABI styles: the typed-C `amm_client_ffi` (swap primitives, under programs/) and the JSON/wire `amm_client` (new-position flow, under apps/). Fold both into a single `amm_client` crate exposing one JSON C ABI, and delete `programs/amm/client-ffi`.

Rust:
- Re-express the swap path as `api/swap.rs` operations on the existing `call::<T>` dispatch — swap_pair, resolve_pool, swap_plan, program_id — reusing `pair::derive_pair` (no more duplicated PDA derivation) and `risc0_zkvm::serde` for the SwapExactInput words (the same encoding the guest decodes). The account list and signer flags stay byte-identical to the old typed path.
- Generate a single header (`include/amm_client.h`) covering all ops via cbindgen; bump cbindgen 0.27 -> 0.28 for `#[unsafe(no_mangle)]` support.

C++:
- Extend the `AmmClient` wrapper with the four swap ops.
- Add `SwapRuntime` (mirrors `NewPositionRuntime`): reads accounts through the wallet, drives the swap ops, submits the transaction.
- `AmmUiBackend` swap methods now delegate to `SwapRuntime`, dropping ~390 lines of typed-FFI and byte-twiddling. `program_id` becomes a JSON op, and the swap clock is derived via `derive_pair` (clock_core::CLOCK_01) instead of a hardcoded base58 literal — same account, verified.


apps/amm/client/include/amm_client.h
```diff
@@ -1,20 +1,49 @@
...
 void amm_free(char *value);
 
 #ifdef __cplusplus
-}
-#endif
+}  // extern "C"
+#endif  // __cplusplus
 
-#endif
+#endif  /* AMM_CLIENT_H */
```
0x-r4bbit, line 49: THe changes in this file are here because the headers are not autogenerated (this was also done in the amm_client ffi from the swap fuctionality PR)

PR #237, opened 2026-07-29
test(amm-ui): add swap UI test

gravityblast:


apps/amm/amm-tokens.json.example
```diff
@@ -0,0 +1,16 @@
...
+    "decimals": 18
+  },
+  {
+    "symbol": "TKB",
+    "name": "Token B",
+    "definitionId": "REPLACE_WITH_TOKEN_B_DEFINITION_ID",
+    "holding": "REPLACE_WITH_YOUR_TOKEN_B_HOLDING_ACCOUNT",
+    "decimals": 18
+  }
+]
```
0x-r4bbit, line 16: We probably want to make this an app-config.json or something like that, I could imagine adding other things here that the app wants to show by default (like a list of pools etc)

apps/amm/amm-tokens.json.example
```diff
@@ -0,0 +1,16 @@
...
+    "definitionId": "REPLACE_WITH_TOKEN_A_DEFINITION_ID",
+    "holding": "REPLACE_WITH_YOUR_TOKEN_A_HOLDING_ACCOUNT",
+    "decimals": 18
+  },
+  {
+    "symbol": "TKB",
+    "name": "Token B",
+    "definitionId": "REPLACE_WITH_TOKEN_B_DEFINITION_ID",
+    "holding": "REPLACE_WITH_YOUR_TOKEN_B_HOLDING_ACCOUNT",
+    "decimals": 18
```
0x-r4bbit, line 14: The decimals don't exist in the program. Is this used anywhere? Otherwise we can get rid off it.

0x-r4bbit (commented):
I tried running this one headlessly via:

```
nix build .#integration-test -L
```

But it fails.

Anything else missing for the headless run?

PR #249, opened 2026-08-04
feat(modules/amm): add swap_exact_in_quote op and module swapExactInQ…

0x-r4bbit:
…uote

Server-side SwapExactInput preview. The swap_exact_in_quote FFI op orients the pool reserves to the requested in/out direction, prices via the shared amm_core::swap_exact_in_amounts (so expectedOut matches the chain), and derives the slippage floor: { expectedOutRaw, minReceivedRaw, priceImpactBps }. no_pool is returned as an error; pool metadata (reserves/fee) comes from resolve_pool, so it isn't echoed. Read-only — no quoteHash; the on-chain min_amount_out is the real guard.

The module swapExactInQuote(tokenIn, tokenOut, amountIn, slippageBps) method derives the pool via the config-free pool_id op, reads it, and wraps the op in the { status, error, ... } envelope; call() now surfaces the op error code so no_pool propagates.

The QML swap view still uses the old path, so nothing breaks. Will make the QML consume it in a follow-up.

### DIFF
--- a/modules/amm/ffi/src/api/swap.rs
+++ b/modules/amm/ffi/src/api/swap.rs
@@ -3,14 +3,18 @@
 //! new-position ops: pure functions returning JSON `Value`, PDAs reused from
 //! `pair::derive_pair` so the swap path never re-derives seeds.
 
-use amm_core::{compute_pool_pda, PoolDefinition};
+use amm_core::{
+    compute_pool_pda, mul_div_floor, price_impact_bps, swap_exact_in_amounts, PoolDefinition,
+    FEE_BPS_DENOMINATOR,
+};
 use nssa_core::account::AccountId;
 use risc0_binfmt::ProgramBinary;
 use serde_json::{json, Value};
 
 use super::{
     pair::{derive_pair, is_canonical_pair},
-    PoolIdRequest, ProgramIdRequest, ResolvePoolRequest, SwapPairRequest, SwapPlanRequest,
+    PoolIdRequest, ProgramIdRequest, ResolvePoolRequest, SwapExactInQuoteRequest, SwapPairRequest,
+    SwapPlanRequest,
 };
 use crate::account::{
     account_id_from_hex, account_id_hex, decode_account, parse_program_id, program_id_bytes,
@@ -82,8 +86,7 @@ pub(super) fn resolve_pool(request: ResolvePoolRequest) -> Result<Value, String>
     if pool.liquidity_pool_supply == 0 {
         return Ok(json!({ "exists": false }));
     }
-    let fee_bps =
-        u32::try_from(pool.fees).map_err(|_| String::from("pool fee tier exceeds u32"))?;
+    let fee_bps = u32::try_from(pool.fees).map_err(|_| String::from("invalid_fee_tier"))?;
     Ok(json!({
         "exists": true,
         "defAHex": account_id_hex(pool.definition_token_a_id),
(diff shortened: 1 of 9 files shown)


modules/amm/ffi/src/api/swap.rs
```diff
@@ -109,6 +112,73 @@ pub(super) fn pool_id(request: PoolIdRequest) -> Result<Value, String> {
...
+/// `min_amount_out` is the real guard. Returns `Err("no_pool")` when the pool is
+/// absent / undecodable / has no liquidity; pool metadata (reserves, fee) comes
+/// from `resolve_pool`, so it isn't echoed here.
+pub(super) fn swap_exact_in_quote(request: SwapExactInQuoteRequest) -> Result<Value, String> {
+    let token_in = account_id_from_hex(&request.token_in_id, "token in id")?;
+    let token_out = account_id_from_hex(&request.token_out_id, "token out id")?;
+    if token_in == token_out {
+        return Err(String::from("swap requires two distinct tokens"));
+    }
+    let amount_in = parse_u128(&request.amount_in_raw, "amountInRaw")?;
```
3esmit, line 132: `amount_in_raw` is parsed but never checked for a positive effective input. For
a valid 30-bps pool, `amount_in_raw = "0"` (and also `"1"`, which fee-rounds to
zero) makes `swap_exact_in_amounts` return `(0, 0)`, so this operation returns a
successful quote with `expectedOutRaw = "0"`. The submission path reaches
`amm_program::swap::swap_logic`, which rejects the same input at its
`effective_amount_in != 0` guard before any transfer. The preview therefore
reports a quote that cannot execute. Reject zero or fee-rounded inputs here and
add a regression test for the same boundary.

modules/amm/ffi/src/api/swap.rs
```diff
@@ -109,6 +112,73 @@ pub(super) fn pool_id(request: PoolIdRequest) -> Result<Value, String> {
...
+    }
+
+    // Exact on-chain pricing (shared with amm_program::swap), then the slippage floor.
+    let (_, expected_out) = swap_exact_in_amounts(amount_in, reserve_in, reserve_out, pool.fees);
+    let slippage_complement = FEE_BPS_DENOMINATOR - u128::from(request.slippage_bps);
+    let min_received = mul_div_floor(expected_out, slippage_complement, FEE_BPS_DENOMINATOR);
+
+    // Price impact (display): how far the realized output falls below the naive
+    // spot valuation, in bps (fee + curve movement combined).
+    let spot_out = mul_div_floor(reserve_out, amount_in, reserve_in);
```
3esmit: The display-only `spot_out` calculation can panic for a valid pool even when
the actual quote is representable. For `reserve_in = 1`,
`reserve_out = u128::MAX`, fee `30`, and `amount_in_raw = "2"`, the shared
on-chain formula uses effective input `1` and returns
`floor(u128::MAX / 2)`. This line instead computes
`floor(u128::MAX * 2 / 1)`, whose quotient exceeds `u128`; `mul_div_floor`
panics while downcasting it, and the FFI returns `internal panic` instead of a
quote. The existing nonzero-reserve guard does not exclude this state. Compute
price impact without materializing an out-of-range `spot_out`, or use a
non-panicking widened/clamped calculation, and add a regression test.

3esmit (dismissed):
Two actionable findings in the exact-input quote path:

1. Zero and fee-rounded-to-zero inputs are accepted even though the guest
   rejects them. See `modules/amm/ffi/src/api/swap.rs:129`.
2. A valid extreme reserve ratio can panic the display-only price-impact
   calculation. See `modules/amm/ffi/src/api/swap.rs:163`.

Other changed paths correctly reuse the on-chain formula, orient reserves by
token direction, and propagate `no_pool` through the C++ module envelope.

PR #252, opened 2026-08-04
feat(modules/amm): add swap_exact_out_quote op and module swapExactOu…

0x-r4bbit:
…tQuote

Server-side SwapExactOutput preview. The swap_exact_out_quote FFI op orients the pool reserves to the requested in/out direction, prices via the shared amm_core::swap_exact_out_amounts (so requiredIn matches the chain), and derives the slippage ceiling: { requiredInRaw, maxInRaw, priceImpactBps }. no_pool and output_exceeds_liquidity (amount_out >= reserve) are returned as errors. Read-only — no quoteHash; the on-chain max_amount_in is the real guard.

The module swapExactOutQuote(tokenIn, tokenOut, amountOut, slippageBps) method derives the pool via the config-free pool_id op, reads it, and wraps the op in the { status, error, ... } envelope. Mirrors swapExactInQuote.

Not yet consumed by the QML swap view, so nothing breaks. Lets the buy field stay editable when the SwapCard is rewired in a follow-up.

### DIFF
--- a/modules/amm/ffi/src/api/swap.rs
+++ b/modules/amm/ffi/src/api/swap.rs
@@ -4,8 +4,8 @@
 //! `pair::derive_pair` so the swap path never re-derives seeds.
 
 use amm_core::{
-    compute_pool_pda, mul_div_floor, price_impact_bps, swap_exact_in_amounts, PoolDefinition,
-    FEE_BPS_DENOMINATOR,
+    compute_pool_pda, mul_div_ceil, mul_div_floor, price_impact_bps, swap_exact_in_amounts,
+    swap_exact_out_amounts, PoolDefinition, FEE_BPS_DENOMINATOR,
 };
 use nssa_core::account::AccountId;
 use risc0_binfmt::ProgramBinary;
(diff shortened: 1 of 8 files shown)


modules/amm/ffi/src/api/swap.rs
```diff
@@ -179,6 +180,76 @@ pub(super) fn swap_exact_in_quote(request: SwapExactInQuoteRequest) -> Result<Va
...
+/// `quoteHash`; the on-chain `max_amount_in` is the real guard. Errors: `no_pool`
+/// (absent / no liquidity) or `output_exceeds_liquidity` (asking for at least the
+/// whole reserve). Pool metadata (reserves/fee) comes from `resolve_pool`.
+pub(super) fn swap_exact_out_quote(request: SwapExactOutQuoteRequest) -> Result<Value, String> {
+    let token_in = account_id_from_hex(&request.token_in_id, "token in id")?;
+    let token_out = account_id_from_hex(&request.token_out_id, "token out id")?;
+    if token_in == token_out {
+        return Err(String::from("swap requires two distinct tokens"));
+    }
+    let amount_out = parse_u128(&request.amount_out_raw, "amountOutRaw")?;
```
3esmit, line 200: `amount_out_raw` is parsed but zero is accepted. With a valid pool and
`amount_out_raw = "0"`, `swap_exact_out_amounts` returns `Some((0, 0))`, and
the quote succeeds with `requiredInRaw = "0"` and `maxInRaw = "0"`. The actual
submission path reaches `amm_program::swap::exact_output_swap_logic`, which
rejects zero at its `exact_amount_out != 0` guard before calling the shared
helper. This makes the exact-output preview claim a usable quote for an input
the guest cannot execute. Reject zero output here and add a regression test.

3esmit (dismissed):
the exact-output quote accepts `amountOutRaw = "0"`,
while the guest requires a nonzero exact output. See the inline comment at
`modules/amm/ffi/src/api/swap.rs:195`. The remaining added path correctly
reuses the shared inverse formula, handles output at or above the reserve, and
propagates the result through the C++ module envelope.

PR #255, opened 2026-08-04
feat(modules/amm): add swap_exact_out_plan op and module swapExactOutput

0x-r4bbit:
Server-side SwapExactOutput submission, mirroring the exact-in path. The swap_exact_out_plan FFI op resolves the pool via the shared derive_pair, encodes amm_core::Instruction::SwapExactOutput { exact_amount_out, max_amount_in, deadline }, and returns the same four tx-submission fields as swap_exact_in_plan (programId, accountIds, signingRequirements, instruction) in the identical fixed 8-account IDL order — only the user's input holding signs. Recoverable domain failures (same_token_pair, config_unavailable) go through the FFI envelope as Err. C export amm_swap_exact_out_plan (cbindgen header regenerated).

The module swapExactOutput(defA, defB, userIn, userOut, amountOut, maxIn, deadline) method reads the AMM config, calls the op, and submits the plan via send_generic_public_transaction, returning the tx hash (empty string on failure). Same argument conventions and I/O-free split as swapExactInput.

Not yet wired into AmmUiBackend/QML — the SwapCard rewire is a follow-up. The existing swap UI path is untouched, so nothing breaks.

### DIFF
--- a/modules/amm/ffi/src/api/swap.rs
+++ b/modules/amm/ffi/src/api/swap.rs
@@ -14,7 +14,7 @@ use serde_json::{json, Value};
 use super::{
     pair::{derive_pair, is_canonical_pair},
     PoolIdRequest, ProgramIdRequest, ResolvePoolRequest, SwapExactInPlanRequest,
-    SwapExactInQuoteRequest, SwapExactOutQuoteRequest, SwapPairRequest,
+    SwapExactInQuoteRequest, SwapExactOutPlanRequest, SwapExactOutQuoteRequest, SwapPairRequest,
 };
 use crate::account::{
     account_id_from_hex, account_id_hex, decode_account, parse_program_id, program_id_bytes,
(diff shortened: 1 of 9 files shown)


modules/amm/ffi/src/api/swap.rs
```diff
@@ -336,6 +336,71 @@ pub(super) fn swap_exact_in_plan(request: SwapExactInPlanRequest) -> Result<Valu
...
+    // Take the vaults from the pool's STORED ids, in its `def_a`/`def_b` (creation)
+    // order — the guest asserts the provided vaults against `pool.vault_a_id` /
+    // `vault_b_id`, which needn't match `derive_pair`'s canonical order for a
+    // non-canonically created pool. (`pair.pool`/`current_tick`/`clock` are
+    // order-independent, so they still come from `derive_pair`.)
+    let Some(pool) = hex::decode(&request.pool_data)
+        .ok()
+        .and_then(|bytes| borsh::from_slice::<PoolDefinition>(&bytes).ok())
+    else {
+        return Err(String::from("no_pool"));
```
3esmit, line 369: `swap_exact_out_plan` returns `Err("no_pool")` when `pool_data` is missing or undecodable, whereas `swap_exact_in_plan` returns `Ok({ "status": "error", "code": "no_pool" })`.

The doc comment says both ops share the envelope contract. `Err("no_pool")` is cleaner because `AmmModuleImpl` checks `planResult.ok` and fails closed immediately instead of building a tx with empty account/instruction vectors.

Please update the doc comment to mention `Err("no_pool")`, and consider aligning `swap_exact_in_plan` to `Err("no_pool")` in a follow-up.

PR #268, opened 2026-08-10
feat(apps/amm): pick the token account per swap side

0x-r4bbit:
Add a per-slot account selector to the swap card's sell/buy inputs, sourced from the wallet's token holdings, so a user holding a token in several accounts chooses which funds each side. A single holding auto-selects.

- SwapPage fetches backend.tokenHoldings(), refetching when the wallet opens
- TokenInput embeds ProgramAccountSelector (half width, right-aligned) in a full-width row below the amount/token row; exposes it as selectedHoldingId
- SwapCard drives swapExactInput/Output with the selected holdings and blocks submit ("Select token accounts") until both are chosen
- ProgramAccountSelector gains showWhenSingle + textAlignment opt-ins

### DIFF
--- a/apps/amm/qml/components/swap/SwapCard.qml
+++ b/apps/amm/qml/components/swap/SwapCard.qml
@@ -17,6 +17,12 @@ Rectangle {
     // Real backend replica (logos.module("amm_ui")), wired from SwapPage.
     property var backend: null
 
+    // The wallet's token holdings (backend.tokenHoldings()), fed to each slot's
+    // account selector; the chosen input/output holding ids drive the submit.
+    property var holdings: []
+    readonly property string sellHolding: sellTokenInput.selectedHoldingId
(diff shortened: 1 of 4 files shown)


3esmit (dismissed):
SwapPage.qml:24-30, 32-43

refreshHoldings starts an asynchronous tokenHoldings request and assigns every
callback directly to root.holdings. Both onBackendChanged and
onIsWalletOpenChanged call it, so requests can overlap. For example, a request
started while the wallet is closed can resolve with an empty list after the
request started by opening the wallet has already returned the real holdings.
That late callback overwrites the current list; a fast wallet switch has the
same failure mode with holdings from the previous wallet. SwapCard then loses
the selected accounts and its canSubmit gate remains false even though the
current wallet has valid holdings.

Capture a refresh generation (or the wallet/backend snapshot) and ignore
callbacks that are no longer current.

PR #269, opened 2026-08-10
test(apps/amm): select the funding account before submitting a swap

0x-r4bbit:
canSubmit now requires a chosen holding on both swap sides, so the swap e2e test must select them first. The selector auto-selects a single holding, but pick it explicitly (robust to multi-account wallets) via a new selectAccount step that waits for the holdings to load and selects the first match.

Expose each side's selector via selectorObjectName (swapSell/BuyAccountSelector) and surface sellHolding/buyHolding in the failure diagnostics.

### DIFF
--- a/apps/amm/tests/swap.mjs
+++ b/apps/amm/tests/swap.mjs
@@ -84,6 +84,38 @@ async function pickToken(app, index) {
   return symbol;
 }
 
+// Pick the funding account for a swap side. The selector auto-selects when the
+// wallet holds the token in exactly one account, but make the choice explicit (and robust
+// to multi-account wallets with zero-balance holdings): wait for the holdings to populate,
+// then select the highest-balance match. canSubmit now requires BOTH sides' holdings to be
+// set, so a swap can't be submitted until this runs.
+async function selectAccount(app, selectorObjectName) {
+  const id = await idByObjectName(app, selectorObjectName);
+  await app.waitFor(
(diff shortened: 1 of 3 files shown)


apps/amm/tests/swap.mjs
```diff
@@ -84,6 +84,29 @@ async function pickToken(app, index) {
   return symbol;
 }
 
+// Pick the funding account for a swap side. The selector auto-selects when the
+// wallet holds the token in exactly one account, but make the choice explicit
+// (and robust to multi-account wallets): wait for the holdings to populate, then
+// select the first match. canSubmit now requires BOTH sides' holdings to be set,
+// so a swap can't be submitted until this runs.
+async function selectAccount(app, selectorObjectName) {
```
3esmit, line 92: selectAccount waits on hasFunds, but ProgramAccountSelector defines hasFunds
as matchingAccounts.length > 0 and does not inspect balanceRaw. The
tokenHoldings response includes zero-balance fungible holdings. The subsequent
evaluate call always selects matchingAccounts[0].

With two holdings for the selected token ordered as balanceRaw 0 followed by
balanceRaw 100, the helper selects the empty account, then the test submits
SELL_AMOUNT 100 through SwapCard. The transfer cannot debit that holding, so
the test fails even though a usable matching account exists. Select a
non-zero holding with enough balance for the requested amount, or make the
fixture exclude empty holdings.

3esmit (dismissed):

PR #271, opened 2026-08-10
feat(modules/amm): add the add-liquidity API and quoting

0x-r4bbit:
Introduce the add-liquidity vertical mirroring the swap + createPool patterns, for depositing into an existing pool via the AddLiquidity instruction.

FFI (amm_ffi):
- add_liquidity_quote — decode poolData, orient the caller's max amounts to the pool's canonical order, run the guest's exact ideal->actual->delta_lp math, and return { amountARaw, amountBRaw, expectedLpRaw, priceRaw } in display order. Slippage-free like create's quote; the min-LP floor is applied at submit.
- add_liquidity_plan — canonicalize (token, max-amount, holding) as one unit, take vaults + LP definition from poolData, emit the 10-account AddLiquidity order (only the user holdings a/b/LP sign). Takes minLpRaw directly, like swap_exact_in_plan takes min_out.

Shared with createPool: extract canonical_triples (the pair/amount/holding canonical swap) and plan_response (the tx-submission envelope); both the create and add plans now use them.

### DIFF
--- a/modules/amm/ffi/src/api/liquidity.rs
+++ b/modules/amm/ffi/src/api/liquidity.rs
@@ -11,12 +11,16 @@
 //! existence before calling; a stale preview or a raced create just reverts on the
 //! guest's `assert pool uninitialized`.
 
-use amm_core::{isqrt_product, spot_price_q64_64, MINIMUM_LIQUIDITY};
+use amm_core::{
+    isqrt_product, mul_div_floor, spot_price_q64_64, PoolDefinition, MINIMUM_LIQUIDITY,
+};
+use nssa_core::account::AccountId;
 use serde_json::{json, Value};
 
 use super::{
     pair::{derive_pair, is_canonical_pair},
-    CreatePoolPlanRequest, LiquidityQuoteRequest,
+    AddLiquidityPlanRequest, AddLiquidityQuoteRequest, CreatePoolPlanRequest,
+    LiquidityQuoteRequest,
 };
 use crate::account::{account_id_from_hex, account_id_hex, parse_program_id};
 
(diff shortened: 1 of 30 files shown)


modules/amm/ffi/src/api/liquidity.rs
```diff
@@ -152,14 +186,163 @@ pub(super) fn create_pool_plan(request: CreatePoolPlanRequest) -> Result<Value,
...
+    }
+    let holding_a = account_id_from_hex(&request.user_holding_a_id, "user holding A id")?;
+    let holding_b = account_id_from_hex(&request.user_holding_b_id, "user holding B id")?;
+    let user_lp = account_id_from_hex(&request.user_holding_lp_id, "user LP holding id")?;
+
+    let max_a = positive_amount(Some(&request.max_amount_a_raw))?;
+    let max_b = positive_amount(Some(&request.max_amount_b_raw))?;
+    let min_lp = positive_amount(Some(&request.min_lp_raw))?;
+    let deadline = parse_u64(&request.deadline_ms, "deadlineMs")?;
+
```
3esmit, line 305: The plan canonicalizes token, amount, and holding with canonical_triples, then
places those holdings beside pool.vault_a_id and pool.vault_b_id from
pool_data. Those vaults are in the pool's stored definition order, which is
not required to be canonical. The guest's AddLiquidity entry point passes
user_holding_a to the token transfer into vault_a; the token program rejects
that transfer when their definition IDs differ.

Concrete input: request token A is 0x11..., token B is 0x22..., the caller
holdings match those display tokens, and pool_data stores definition_token_a
as 0x11... and definition_token_b as 0x22... (a non-canonical stored pair).
canonical_triples swaps the user holdings because 0x22... is the canonical
first token, then the plan sends the token-B holding to vault_a for token A.
The transaction reaches the chained transfer and fails with a sender/recipient
definition mismatch.

Orient the amount/holding pairs against pool.definition_token_a_id and
pool.definition_token_b_id, and reject a pair mismatch before encoding the
plan.

3esmit (changes requested):

modules/amm/ffi/src/api/liquidity.rs
0x-r4bbit: Great find!

PR #278, opened 2026-08-10
feat(amm): drive the create-pool liquidity preview from liquidityQuote

0x-r4bbit:
Both liquidity branches now quote through the lean composable ops. The create path joins the add path (already on addLiquidityQuote) by reworking liquidity_quote into a dual-mode create quote and wiring the form to it via the resolvePool pool read. quoteNewPosition/amm_quote are no longer reached from the UI.

FFI (modules/amm/ffi):
- liquidity_quote is dual-mode: price-only (initialPriceRealRaw, no amounts) returns the minimum opening deposit via minimum_opening_pair; supplied amounts return the actual deposit with the price derived from them. Emits actual/minimum amounts, expectedLp, lockedLp and the Q64.64 price.
- LiquidityQuoteRequest gains initial_price_real_raw (Option<String>, needed only in price-only mode).

Module (modules/amm/src):
- liquidityQuote forwards initialPriceRealRaw to the op.

UI (apps/amm/qml):
- Route create-vs-add on the pool read (resolvePool -> poolExists); create quotes via liquidityQuote, assembled into the missing-pool shape the form already consumes.
- poolStatus moves off the quote onto the flow's poolExists; the form derives activePool/missingPool from it. Trim the vestigial quote fields (canSubmit, requiresFreshLp, warnings, errors[], accountPreview, the "Pool" row) and drop the account-plan panel for parity with the swap view.
- Fix a real bug: a pair change now resets poolExists (resetPoolExistence) so resetPairDraft re-resolves the pool like a fresh selection. Otherwise an active pool kept stale (cleared) reserves with no re-quote, and the deposit ratio-fill silently no-op'd.

Tests (apps/amm/tests):
- Read activePool instead of the removed poolStatus. The add test waits for the reset's active-pool quote to settle (reserves reloaded) before the ratio-fill; the create test resets the draft to clear leftover cross-run amounts and the stale submitted transactionId.

### DIFF
--- a/apps/amm/qml/components/liquidity/NewPositionForm.qml
+++ b/apps/amm/qml/components/liquidity/NewPositionForm.qml
@@ -27,6 +27,8 @@ AmmActionCard {
     property var holdings: []
     readonly property string selectedHoldingAId: tokenAInput.selectedHoldingId
     readonly property string selectedHoldingBId: tokenBInput.selectedHoldingId
+    readonly property string selectedBalanceARaw: tokenAInput.selectedBalanceRaw
+    readonly property string selectedBalanceBRaw: tokenBInput.selectedBalanceRaw
     property string selectedTokenAId: ""
     property string selectedTokenBId: ""
     property int selectedFeeBps: 30
@@ -43,7 +45,6 @@ AmmActionCard {
     property string tokenResolutionError: ""
     property string tokenResolutionErrorSide: ""
     property string tokenResolutionMessage: ""
-    property string confirmedPoolStatus: ""
     property var activePoolQuote: ({})
     property string headingText: qsTr("New position")
     property string headingDetail: ""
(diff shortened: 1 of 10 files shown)


apps/amm/qml/components/liquidity/NewPositionForm.qml
```diff
@@ -100,11 +100,12 @@ AmmActionCard {
...
     // Both deposit amounts must be present. For create the quote auto-fills them from the
     // opening deposit; for add the user enters one and the other ratio-fills. Without this
-    // the active-pool probe quote (sent on pair-select with simulated amounts) reports
-    // canSubmit, wrongly enabling the CTA before any amount is entered.
+    // an active-pool probe quote (sent on pair-select with simulated amounts) would enable
+    // the CTA before any amount is entered.
     readonly property bool hasDepositAmounts: root.amountA.length > 0 && root.amountB.length > 0
+    // A successful lean quote is submittable — funding is gated below (holdings + amounts),
+    // not by a quote-side canSubmit flag.
     readonly property bool canConfirm: root.quotePayload.status === "ok"
```
3esmit, line 110: canConfirm now requires only an ok quote, selected holdings, non-empty
amounts, and loading state. The previous quote-side funding/canSubmit gate is
gone, while the lean liquidityQuote and addLiquidityQuote APIs do not compare
the requested amounts with the selected holdings' balanceRaw. TokenAmountInput
only displays balance, and selectedBalanceRaw is not consumed by the form.

With a selected holding whose balance is below the entered amount, the lean
quote still returns status ok and canConfirm enables the CTA. The subsequent
createPool or addLiquidity transfer then fails for insufficient balance. Add
a balance-versus-amount check before enabling confirmation, or restore an
operation that reports funding errors.

3esmit (changes requested):
pendingQuoteRequest is reused after a submit failure. A create-pool request
contains amountARaw, amountBRaw, and initialPriceRealRaw, but the add branch
reads maxAmountARaw and maxAmountBRaw from that same request. If another
transaction creates the pool after the create quote but before the failed
submit callback retries, resolvePool returns an active pool and
requestAddQuote sends undefined max amounts. The addLiquidityQuote module path
then returns bad_amount, so the automatic retry fails with a schema error
instead of rebuilding an add request from the current draft.

Keep branch-specific request builders or discard the create request when the
pool branch changes before retrying.

